PR #61, opened 2015-08-02
Dev

NikCharlebois:
Added the new optional CentralAdministrationPort parameter for users to specify a custom port to run Central Administration on.


BrianFarnhill:
Nik can you let me know why the invoke-command calls have been removed? We need to run these as a specific user rather than the local system account.

NikCharlebois:
this was an old version of a test that somehow got submitted as part of the pull request. My bad. 'll resubmit the pull request properly.

Issue #43, opened 2015-06-22
Extend xSPWebApplication resource to include "workflow settings"


BrianFarnhill:
Extend the xSPWebAppliation resource top include the follow settings (seen in central admin under the "workflow settings" dialog)

 - Enable user defined workflows
 - Allow internal users with no site access to get tasks
 - Allow external users to get tasks with a copy of the document

camiloborges:
hello. which ones? 
I think you are referring to the following properties.

Allow internal users with no site access to get tasks
     EmailToNoPermissionWorkflowParticipantsEnabled


Allow external users to get tasks with a copy of the document
     ExternalWorkflowParticipantsEnabled
     

Enable user defined workflows:
    UserDefinedWorkflowsEnabled 

What about these ones?
     AllowDesigner
     AllowCreateDeclarativeWorkflow 
     AllowSaveDeclarativeWorkflowAsTemplate
     AllowSavePublishDeclarativeWorkflow 

I am not sure though.

BrianFarnhill:
The initial ask was to mimic the settings on the workflow settings screen in central admin (when you select a web app then choose workflow settings from that drop down item in the ribbon), but happy to discuss others that aren't on there if you think they are relevant.

BrianFarnhill:
This is part of PR #96 and will be closed when this is merged to dev

BrianFarnhill:
This has been merged to dev in PR #96 and will be released in v0.8 - closing this issue

PR #63, opened 2015-08-03
Added Workflow Settings  - Fix for Issue #43

NikCharlebois:


BrianFarnhill:
I'll pump this one through a quick test this week, I also want to review the tests as well to see if we should update the scenarios in the test methods for the web app object to account for these new properties as well. If you could have a quick look and see if you can add to that first as well that would be great.

NikCharlebois:
Thanks Brian. Are you talking about the xSharePoint.xSPWebApplication.Tests.ps1 script or the actual Test-TargetResource function of the Resource?

BrianFarnhill:
Sorry should have specified - was talking about the tests.ps1 file. I've been meaning to review this resource to look at what happens with optional parameters where they aren't specified in the configuration (for example, I manually turn on one of the workflow options, but don't specify a setting for it in the DSC - does the DSC default to false as the default for a bool value, and then turn it off for me or does it ignore it and leave the current setting). This is a wider issue across some of the other resources we have but I wanted to throw it out there to discuss seeing as we are adding more of these types of values so we can get it sorted before we get it out of hand.

NikCharlebois:
I've got us covered for this :). I will be sending another pull request� later today to add the throttling params to web application, and I've added code to handle the case where optional params are not passed to $PSBoud

Sent from my BlackBerry 10 smartphone on the Rogers network.
From: Brian Farnhill
Sent: Wednesday, August 5, 2015 5:53 AM
To: PowerShell/xSharePoint
Reply To: PowerShell/xSharePoint
Cc: Charlebois-Laprade, Nikolas
Subject: Re: [xSharePoint] Added Workflow Settings - Fix for Issue #43 (#63)



Sorry should have specified - was talking about the tests.ps1 file. I've been meaning to review this resource to look at what happens with optional parameters where they aren't specified in the configuration (for example, I manually turn on one of the workflow options, but don't specify a setting for it in the DSC - does the DSC default to false as the default for a bool value, and then turn it off for me or does it ignore it and leave the current setting). This is a wider issue across some of the other resources we have but I wanted to throw it out there to discuss seeing as we are adding more of these types of values so we can get it sorted before we get it out of hand.

�

PR #64, opened 2015-08-06
Supports WebApplication Resource Throttling, Workflow Settings, and bounds optional parameters to PSBound

NikCharlebois:


BrianFarnhill:
Thanks Nik, I'll review and functional test this then merge when that's done.

BrianFarnhill:
Hey Nik, are you able to explain the logic behind the Add-Variable function you have added in MSFT_xSPWebApplication.psm1. I'm not sure I'm following why it has been created, I can see you using it to pump values in to the PSBoundParameters variable, but that array should already contain values that have been specifically bound to the function call.

NikCharlebois:
Hi Brian, maybe the issue has been fixed in latest wmf versions, but at some point only mandatory params were "psbound". Optional ones were not properly passed, which is why I am forcing the bound operation on all variables, with their defaults if they have not been provided with another value.

Sent from my BlackBerry 10 smartphone on the Rogers network.
From: Brian Farnhill
Sent: Monday, August 17, 2015 5:46 PM
To: PowerShell/xSharePoint
Reply To: PowerShell/xSharePoint
Cc: Charlebois-Laprade, Nikolas
Subject: Re: [xSharePoint] Supports WebApplication Resource Throttling, Workflow Settings, and bounds optional parameters to PSBound (#64)



Hey Nik, are you able to explain the logic behind the Add-Variable function you have added in MSFT_xSPWebApplication.psm1. I'm not sure I'm following why it has been created, I can see you using it to pump values in to the PSBoundParameters variable, but that array should already contain values that have been specifically bound to the function call.

�

BrianFarnhill:
OK, I've just done a few quick tests and it appears that optional params come over if they are specified, and not included if they are not specified. Are you able to replicate that issue in any specific environment? If not I'm inclined to take this out of the script to reduce the overall complexity.

BrianFarnhill:
Hey Nik, I think I've come across the scenario you were talking about with the optional parameters actually - basically we need to have something that adds the optional params to PSBoundParameters where there is a default value so that when we pass the PSBoundParams as an argument to the Invoke-Command cmdlet it has the default value in there for it to act upon as that scope wont have the existing variable - that sound about right? 

If so - we need to come up with an approach that we can keep simple and then more than likely drop in to the utilities scripts as it will need to be used in a bunch of different places. Leave this with me as I need to fix it as part of some other work I'm doing, but once I've punched the generic utilities function in for this can we update this PR to use that and then merge it in?

NikCharlebois:
Yep, you've got it, that's exactly the issue. Yes, it needs to be centrally located in the utility module for sure! I am back in business after two weeks vacations. I will be talking about your DSC resource at SPTechCon this week in Boston :) I will make sure to do all the publicity I can :)M so you will be taking care of moving it into the utility? I can do it if you'd like, I've got some cycles this week. Just let me know

Nik

Sent from my BlackBerry 10 smartphone on the Rogers network.
From: Brian Farnhill
Sent: Sunday, August 23, 2015 7:27 PM
To: PowerShell/xSharePoint
Reply To: PowerShell/xSharePoint
Cc: Charlebois-Laprade, Nikolas
Subject: Re: [xSharePoint] Supports WebApplication Resource Throttling, Workflow Settings, and bounds optional parameters to PSBound (#64)



Hey Nik, I think I've come across the scenario you were talking about with the optional parameters actually - basically we need to have something that adds the optional params to PSBoundParameters where there is a default value so that when we pass the PSBoundParams as an argument to the Invoke-Command cmdlet it has the default value in there for it to act upon as that scope wont have the existing variable - that sound about right?

If so - we need to come up with an approach that we can keep simple and then more than likely drop in to the utilities scripts as it will need to be used in a bunch of different places. Leave this with me as I need to fix it as part of some other work I'm doing, but once I've punched the generic utilities function in for this can we update this PR to use that and then merge it in?

�

BrianFarnhill:
I've got some time dedicated to work on it this week ahead of my presentation at Ignite New Zealand next week, so I'll pick it up and let you know when it's sorted mate

BrianFarnhill:
OK so from looking at my lab environment, the DSC engine is handling the optional parameters a little differently to a normal PowerShell function. I took your Add-Variable function and tested it and saw that my defaults weren't ever being set, it was always null. I then wrote a slightly different version with some more logging and when the function was called in DSC it was always null - if I manually loaded the module and called it then it would work though - so I think we need to handle the optional parameters differently here. What I found worked was this:

if (-not $PSBoundParameters.ContainsKey("CentralAdministrationPort")) { $PSBoundParameters.Add("CentralAdministrationPort", 9999) }

So instead of relying on the default value of the parameter, I set a default in my own script. What do you think of this approach Nik? Could this work do you think?

NikCharlebois:
�Hey Brian,

I think it could work, but that would make a lot of if statements. I think we need to figure out the root problem here. Why is it that the defaults are not being set in the first place. Reading forums post, everyone seems to assume it doesn't work, but that is not right...PSBound should take them in by default. Talking to you colleagues, are any of them facing the same issues with their DSCs?


From: Brian Farnhill
Sent: Tuesday, August 25, 2015 7:53 PM
To: PowerShell/xSharePoint
Reply To: PowerShell/xSharePoint
Cc: Charlebois-Laprade, Nikolas
Subject: Re: [xSharePoint] Supports WebApplication Resource Throttling, Workflow Settings, and bounds optional parameters to PSBound (#64)



OK so from looking at my lab environment, the DSC engine is handling the optional parameters a little differently to a normal PowerShell function. I took your Add-Variable function and tested it and saw that my defaults weren't ever being set, it was always null. I then wrote a slightly different version with some more logging and when the function was called in DSC it was always null - if I manually loaded the module and called it then it would work though - so I think we need to handle the optional parameters differently here. What I found worked was this:

if (-not $PSBoundParameters.ContainsKey("CentralAdministrationPort")) { $PSBoundParameters.Add("CentralAdministrationPort", 9999) }

So instead of relying on the default value of the parameter, I set a default in my own script. What do you think of this approach Nik? Could this work do you think?

�

BrianFarnhill:
Ill fire off some emails internally to see what I can find, but the ifs might be a workaround in the mean time I think so we can get it working.

NikCharlebois:
Hey, you're the boss :) whatever works is fine by me :p

Sent from my BlackBerry 10 smartphone on the Rogers network.
From: Brian Farnhill
Sent: Tuesday, August 25, 2015 8:19 PM
To: PowerShell/xSharePoint
Reply To: PowerShell/xSharePoint
Cc: Charlebois-Laprade, Nikolas
Subject: Re: [xSharePoint] Supports WebApplication Resource Throttling, Workflow Settings, and bounds optional parameters to PSBound (#64)



Ill fire off some emails internally to see what I can find, but the ifs might be a workaround in the mean time I think so we can get it working.

�

BrianFarnhill:
Hey again Nik,

Just looking over this and testing a little further - I think we need to change the approach to setting the default values on the web app settings. For example, at the moment if I dont want to do the workflow settings or throttling settings through the DSC I will see the web app set them all to default values - I think we need to detect if a value was passed to the configuration then we set it, otherwise we leave it alone rather than set a default. This could also be problematic where the defaults vary between SP2013 and SP2016 as well.

Also can we look at incorporating setting some of the settings where a web app already exists? For example we set the list view threshold to 5000 here, if someone changes it in the UI at the moment we won't set it back to 5000 (which it should), so on the other end of the if ($null -eq $wa) statement in the set method we need to make sure we set all the appropriate settings back to what was supplied to the configuration.

NikCharlebois:
Hi Brian,

I agree with your first comment about us having to figure out a way to leave the defaults alone if they are not provided by the user. If only the defaults were the same between the 2 versions ☺

I will investigate your point about setting back the provided values in the case where a web app already exists. Thanks

Nik Charlebois<http://geds20-sage20.ssc-spc.gc.ca/en/GEDS20/?pgid=015&dn=CN%3DCharleboisLaprade%5C%2C+Nikolas%2COU%3DCCS-CSC%2COU%3DIMCS-GISC%2COU%3DICOB-DGIDPI%2COU%3DCRB-DGIR%2COU%3DLAC-BAC%2CO%3DGC%2CC%3DCA>
i/Gestionnaire – Collaboration et systems corporatifs
a/Manager – Collaboration & Corporate Systems
Bibliothèque et archives Canada / Library & Archives Canada

BrianFarnhill:
@NikCharlebois Hey Nik, I'm going to close this one off and work to role this in through PR #96 - it was a little less leg work to merge and I worked with @camiloborges on the whole "complexity of xSPWebApplication" bit - we've just about finalised this one and well get that merged in soon. Thanks for getting this started though.

PR #76, opened 2015-09-03
Add support for PowerShell 5 PsDscRunAsCredential

BrianFarnhill:


BrianFarnhill:
@caadam - can I get a code review mate?

BrianFarnhill:
Heads up @caadam - I've found an issue in here that I'm working on in some testing, ping me when you're looking at this for the details, but it looks like there is some error coming out where we are loading the SP Snap-In for everything after the first resource that uses it (create or join farm usually)

BrianFarnhill:
Closing this one off while I work to resolve the issue I mentioned above. I believe I have a fix but it involves touching a bunch of other areas first. Will submit a separate PR for the complete functionality as a single unit rather than breaking it up to multiple PRs

PR #80, opened 2015-09-16
Review and repair Get methods on all resources

BrianFarnhill:
Ensure that all resources return an appropriate value out of the get methods. Also added testing to ensure that scripts contain matching schemas to the MOF files (related to ensuring a valid get method but covers all methods in each DSC resource)


BrianFarnhill:
Code review time - @caadam you're up

BrianFarnhill:
Finished patching up the last little bugs in this script - ready to merge post the review being done

caadam:
Good to see code coverage is now north of 85%
Reformatting of code has yielded a good return in terms of readability

BrianFarnhill:
Thanks dude - I've found one last bug in this that im sitting on before I merge this. Will update here with a fix when its found

Issue #89, opened 2015-11-03
Add xSPWebAppBlockedFileTypes resource


BrianFarnhill:
We are looking to add a resource to manage the blocked file types that are used on a web application. The syntax would look like this:

    xSPWebAppBlockedFileTypes BlockFileTypes
    {
        WebAppName = "SharePoint sites"
        BlockedFileTypes = "exe,dll"
        AllowedFileTypes = "config,json"
    }

The concept here is that an administrator can specify a list of file types to ensure are in the blocked file types list for the given web application, but also to specify file types that are to never be in the list (so in the case of them being manually added they can be removed).

camiloborges:
Hello. Just trying to keep things consistent... xSPOutgoingEmailSettings takes WebAppUrl as parameter. This one asks for WebAppName.
Should we have only one way of referring to the web app?

BrianFarnhill:
Nice catch, it should use the URL not the name

BrianFarnhill:
Also I'm considering moving this one to be a property on the existing xSPWebApplication object rather than an individual resource, let me know if you are going to work on this one first.

kilasuit:
I would suggest that this should be as part of the xSPWebApplication resource rather than an individual resource

kilasuit:
To expand on that setting that are able to be applied at the WebApplication level I would be more inclined to group them under the xSPWebApplication resource

camiloborges:
I was half way through it already :P but happy to change.
it makes sense to be just an enhancement for the xSPWebApplication indeed. I just shut the VMs down, will look into it tomorrow

BrianFarnhill:
Agreed - Camilio ping me and we can have a chat about how to move it to the webapp object. I'll also mark this as in progress with you as well :)

BrianFarnhill:
This has been merged to dev in PR #96 and will be released in v0.8 - closing this issue

PR #96, opened 2015-11-16
adding xSPPasswordChangeSettings, xSPOutgoingEmailSettings, and adding blocked file types.

camiloborges:


Modules/xSharePoint/DSCResources/MSFT_xSPOutgoingEmailSettings/MSFT_xSPOutgoingEmailSettings.psm1
```diff
@@ -0,0 +1,86 @@
...
+        [parameter(Mandatory = $true)] [System.String] $CharacterSet
+        
+    )
+
+    Write-Verbose -Message "Retrieving outgoing email settings configuration "
+
+    $result = Invoke-xSharePointCommand -Arguments $PSBoundParameters -ScriptBlock {
+        $params = $args[0]
+        
+        $webApp = $null
```
BrianFarnhill: This $webApp = $null statement is redundant, if you just add -ErrorAction SilentlyContinue to the following line you will get $null out of that if the web app doesn't exist.

Modules/xSharePoint/DSCResources/MSFT_xSPOutgoingEmailSettings/MSFT_xSPOutgoingEmailSettings.psm1
```diff
@@ -0,0 +1,86 @@
...
+        [parameter(Mandatory = $true)] [System.String] $SMTPServer,
+        [parameter(Mandatory = $true)] [System.String] $FromAddress,
+        [parameter(Mandatory = $true)] [System.String] $ReplyToAddress,
+        [parameter(Mandatory = $true)] [System.String] $CharacterSet
+        
+    )
+
+    Write-Verbose -Message "Retrieving outgoing email settings configuration "
+
+    $result = Invoke-xSharePointCommand -Arguments $PSBoundParameters -ScriptBlock {
```
BrianFarnhill: You need to add an InstallAccount parameter (look at pretty much any other resource we have to see how we do it), this is needed to ensure we can support PowerShell 4 where PsDscRunAsCredential is not available.

Modules/xSharePoint/DSCResources/MSFT_xSPPasswordChangeSettings/MSFT_xSPPasswordChangeSettings.psm1
```diff
@@ -0,0 +1,85 @@
...
+        [parameter(Mandatory = $false)][ValidateRange(0,36000)][System.UInt32] $PasswordChangeWaitTimeSeconds,
+        [parameter(Mandatory = $false)][ValidateRange(0,99)][System.UInt32] $NumberOfRetries
+    )
+
+    Write-Verbose -Message "Retrieving farm wide automatic password change settings"
+
+    $result = Invoke-xSharePointCommand -Arguments $PSBoundParameters -ScriptBlock {
+        $params = $args[0]
+        
+        $farm = Get-SPFarm
```
BrianFarnhill: Can you add -ErrorAction Continue to the Get-SPFarm command to ensure the exception doesn't halt the method and the $null value below is returned correctly

Modules/xSharePoint/DSCResources/MSFT_xSPPasswordChangeSettings/MSFT_xSPPasswordChangeSettings.psm1
```diff
@@ -0,0 +1,85 @@
...
+    (
+        [parameter(Mandatory = $true)][System.String] $MailAddress,
+        [parameter(Mandatory = $false)][ValidateRange(0,356)][System.UInt32] $DaysBeforeExpiry,
+        [parameter(Mandatory = $false)][ValidateRange(0,36000)][System.UInt32] $PasswordChangeWaitTimeSeconds,
+        [parameter(Mandatory = $false)][ValidateRange(0,99)][System.UInt32] $NumberOfRetries
+    )
+
+    Write-Verbose -Message "Retrieving farm wide automatic password change settings"
+
+    $result = Invoke-xSharePointCommand -Arguments $PSBoundParameters -ScriptBlock {
```
BrianFarnhill: As with the other resource we need the InstallAccount parameter

Modules/xSharePoint/DSCResources/MSFT_xSPWebApplication/MSFT_xSPWebApplication.psm1
```diff
@@ -92,9 +101,17 @@ function Set-TargetResource
                 $params.Remove("AllowAnonymous") | Out-Null 
                 $params.Add("AllowAnonymousAccess", $true)
             }
-
-            New-SPWebApplication @params
+         
+            $wa = New-SPWebApplication @params
         }
+        write-host "bla"
```
BrianFarnhill: Can you clear up these "bla" statements

Tests/xSharePoint/xSharePoint.xSPOutgoingEmailSettings.Tests.ps1
```diff
@@ -0,0 +1,101 @@
...
+        Import-Module $Global:CurrentSharePointStubModule -WarningAction SilentlyContinue         
+
+
+
+        Context " Web Application isn't available " {
+            Mock Get-SPWebApplication -MockWith  { return $null
+            }
+
+            It "returns null from the get method" {
+                Get-TargetResource @testParams | Should be $null
```
BrianFarnhill: When testing for $null we should use "Should BeNullOrEmpty"

Tests/xSharePoint/xSharePoint.xSPOutgoingEmailSettings.Tests.ps1
```diff
@@ -0,0 +1,101 @@
...
+            It "returns false from the test method" {
+                Test-TargetResource @testParams | Should be $false
+            }
+
+        }
+
+        Context " Properties match" {
+            Mock Get-SPWebApplication { 
+                return @{
+                        Url= "http://sp2013.australiasoutheast.cloudapp.azure.com"
```
BrianFarnhill: Can we make the URLs we mock with more generic examples, like http://sharepoint.contoso.com or something like that?

BrianFarnhill:
Just as a general comment about how we do the blocked file types mate - what I was envisioning (as per the discussion in #89) is that we'll add two properties, one for a list of file types we want to make sure are allowed, and another for a list we want to make sure are blocked. This gets us away from needing to define the entire list (which could be quite long) in every configuration - the theory is that the list will be as it is by default and we'll only add/remove the file types in those two properties as defined in the configuration. This should keep it in a much more manageable form. So we might need to look at reworking that a little bit to line up with that approach. Happy to help you out with that if needed mate, just let me know.

Modules/xSharePoint/DSCResources/MSFT_xSPOutgoingEmailSettings/MSFT_xSPOutgoingEmailSettings.psm1
camiloborges: I thought you had told me not to have the installaccount. Easy enough. I will re-add it:)

BrianFarnhill: No I said to test without it, we need to have it though :)

camiloborges:
All good, I was thinking about how to avoid the overwhelming list. 
Let me have a think about that, but shouldn't be too hard anyway 

Camilo Borges
Mob:0488 056 964

camiloborges:
Hello. 
so, just thinking about the "BlockedFileTypes" and "AllowedFileTypes".

Should the parameter names be updated to "EnsureBlockedFileTypes" and "EnsureAllowedFileTypes" as they are not meant to be the list of all the blocked files?

BrianFarnhill:
Good call Camillo, rename the variables as you suggested.

camiloborges:
Cool. Have a look at that other post that suggests the object model for all the xspwebapp.

Worth mentioning I have scripts for few of the other pending issues(loads on user profile settings). I will try to work on the user profile after I get the ones in your backlog done

ykuijs:
A comment on the blocked file types: Shouldn't we have three different parameters: BlockedFileTypes, EnsureBlockedFileTypes and EnsureAllowedFileTypes?

With the current two, you cannot enforce a specific set of file types. You can only specify types to include and files to exclude. By adding the third parameter (which cannot be used wit the other two), you are abel to do this. For an example, see the xSPFarmAdministrators resource I just created a pull request for. That contains three parameters: Members (enforce a specific list), MembersToInclude (add an admin to the existing list if not present) and MembersToExclude (remove and admin from the list if present).

ykuijs:
An addition: Since the two current properties are used for adding and removing file type to the list. Maybe it is an idea to rename the variables to BlockFileTypesToInclude and BlockedFileTypeToExclude. I think these cover the meaning of the parameters a little better.

kilasuit:
@ykuijs - for configuration updates you would update the BlockedFileTypes or the AllowedFileTypes only and then push (or have it pull) a new configuration to the farm therefore there would be no need for the 

Personally I feel that there should only be a BlockedFileTypes property as there isn't a corresponding AllowedFileTypes in CA so should we be adding this is when it doesn't actually exist as a property to add further unnecessary logic in the resource?

ykuijs:
From a setup/install perspective I can see the need for adding or removing one or a few file types. But from a administration perspective I would like to be able to enforce a specific list of file types. To allow flexibility to the person who is using the resource, we can add both options.

BrianFarnhill:
@ykuijs @kilasuit The thinking behind having the block and the allow list is for administrators who do not wish to tightly manage the entire list of file extensions in their configurations. In this way they can ensure the specific extensions that they care about are managed accordingly. That said, I agree that adding a property for those who wish to lock down the entire list makes sense as well, so I see where Yorick is coming from on this as well. So based on that, I think we can also rename them as Yorick suggested above as well - @camiloborges what do you think, can you update this one?

kilasuit:
But surely if it is not blocked then it is allowed - granted its been some time since I've had to do any CA work but isn't there is only a Blocked file types in the Web Application Settings in CA?

BrianFarnhill:
@kilasuit Yes if it is not blocked then it's allowed, but the scenario we are catering for is this - say an administrator want to allow PowerShell scripts to be stored in their web app, they can add just one property that say allow PS1 files without needing to then specificially list out the few dozen file extensions. So you end up with "whatever the default is, minus PS1 files". Its a more flexible approach for those that only want to care about some file types while still allowing us to give another option for those who want to list out every file type and control it all specifically.

camiloborges:
Hello.
I like the idea, however I feel we are overcooking this.

We are talking about 1 UI property, and we are translating it into 3.

I am always keen to reduce complexity. So, I think we need to have one or the other.

Either all is pretty simple to implement. But I think we need to decide
1) have 1 property that lists all available files
2) have 2 properties to define blocked/allowed
3) make it a list, where we can have
@{Action(replace all/ensureblocked/ensure allowed)
File types("bla","....)
}

I personally think #2 is the simplest yet meaningful approach. Why? I don't care what are all the blocked files, we care about the ones that affect us, and that is the few here and there that need to be either blocked or allowed

Does that make sense?


(Implementing this will be effectively merging first and last commit). There is also a big refactor I am doing in the xspwebapp so now is the moment to really nail down the approach going forward

BrianFarnhill:
@camiloborges - I see where you are coming from and initially I had the same opinion until I spoke to @ykuijs about this one, and we discussed how the out of the box group resource works in that it does the same thing here, takes what is one thing in the UI and uses three properties to give the most flexibility to the admin to control it they way they want to control it. Yes that makes the scripts a little more complex for us, but ultimately isn't that what we are trying to do though? Give a good experience to an admin who is writing the configurations to just tell DSC what they want, and we'll do the complex bits under the hood for them to make it "just work". I still feel that the best approach here is to split it up in the same way that the group resource has done and we can do the leg work in our scripts behind the scenes.

camiloborges:
Sweet. I will included that then. Will take me no time.

What if all of them are provided? Throw exception or just process it In certain order

Camilo Borges
Mob:0488 056 964

BrianFarnhill:
Yea I would check first to make sure they use a combination of properties that is supported, so if the members one is used that the membersToInclude and membersToExclude are both null, and then that if the members one is null that at least one of the other two properties has a value. Then react to that from there. Make sense?

BrianFarnhill:
@camiloborges I'm working on this one now mate, reviewing and making minor tweaks - I'll send a PR to your dev branch with the changes I make through the review. My [work in progress](https://github.com/BrianFarnhill/xSharePoint/tree/merge-pr96) branch will have the latest

BrianFarnhill:
My proposed fixes to the code done by @camiloborges is in a PR for his [dev branch](https://github.com/camiloborges/xSharePoint/pull/1) so that once he looks it over they can be merged in and we can merge this one to dev. This is the last PR I'm waiting on before I fork off to release v0.8

BrianFarnhill:
@camiloborges we're close mate - looks like the new functions aren't being loaded by the module correctly. Ill see if i can replicate this one myself and get you a patch. Let me know if you fix it in the meantime though.

camiloborges:
hey, don't do anything yet. this is one of the things I fixed yesterday. please sync before merging next time :)

camiloborges:
had been fixed on https://github.com/camiloborges/xSharePoint/commit/258e8b06a9ab5b60e5e185f31bc7450fdd7e29bd

BrianFarnhill:
@camiloborges Build is still failing mate - looks like the same issue in the blocked file types module, wanna throw the import module statement in there (and im assuming in the throttling one as well)

camiloborges:
fuck yeah! done!

BrianFarnhill:
@camiloborges never mind - looks like a green light, I'll merge this one in to dev and then fork off for a release branch to touch up anything else that doesn't sit quite right. Thanks for the help on this one dude!

camiloborges:
it seems appVeyor isn't checking builds on my forked copy. will make sure it is so you can't send me a PR that fails validation :p

Issue #93, opened 2015-11-10
Add xSPAntivirusSettings resource


ykuijs:
The xSPAntivirusSettings resource will be used to configure the anti virus farm settings. It will be represented as below:

xSPAntivirusSettings ConfigureAV
{
ScanDocumentsOnUpload=$false
ScanDocumentsOnDownload=$false
AllowUsersToDownloadInfectedDocuments=$true
AttemptToCleanInfectedDocuments=$false
TimeoutDuration=300
NumberOfThreads=5
}

ykuijs:
Working on it!

BrianFarnhill:
Merged in to dev, closing this issue

PR #98, opened 2015-11-19
Added xSPAntivirusSettings and xSPFarmAdministrators

ykuijs:
Added the resources xSPAntivirusSettings (Issue #93) and xSPFarmAdministrators
(still under discussion, currently only a demo)


Modules/xSharePoint/DSCResources/MSFT_xSPFarmAdministrators/MSFT_xSPFarmAdministrators.psm1
```diff
@@ -0,0 +1,224 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [parameter(Mandatory = $true)] [System.String] $Name,
+        [parameter(Mandatory = $false)] [System.String[]] $Members,
+        [parameter(Mandatory = $false)] [System.String[]] $MembersToInclude,
```
kilasuit: MembersToInclude & MembersToExclude are unneccessary to be included as with DSC the LCM will update this on the farm after a new version of the Configuration is Pushed (or pulled) to the servers in the farm.

The reasoning behind as referenced in the other PR would be valid if this was just a custom PS Script that you were running to maintain the configuration but with DSC you should just be updating the minimum required to the configuration and then request to "Make it so" as Jeffery Snover would say

ykuijs:
@kilasuit - I see what you mean and I agree. But what about a situation where you want to add an account to the Farm Administrators group, but do not "care" about any other users in the group? Or if you want to prevent a specific user to be added to the Farm Administrator group, but the rest of the members are flexible. With these two properties you have the flexibility to do that, that is why I added them. You can use DSC to "Make it so" :-)

The code contains a check to prevent you from using Members and any of the other two parameters at the same time.

Modules/xSharePoint/DSCResources/MSFT_xSPFarmAdministrators/MSFT_xSPFarmAdministrators.psm1
ykuijs: @kilasuit - I see what you mean and I agree. But what about a situation where you want to add an account to the Farm Administrators group, but do not "care" about any other users in the group? Or if you want to prevent a specific user to be added to the Farm Administrator group, but the rest of the members are flexible. With these two properties you have the flexibility to do that, that is why I added them. You can use DSC to "Make it so" :-)

The code contains a check to prevent you from using Members and any of the other two parameters at the same time.

Modules/xSharePoint/DSCResources/MSFT_xSPAntivirusSettings/MSFT_xSPAntivirusSettings.psm1
```diff
@@ -0,0 +1,113 @@
...
+        Write-Verbose -Message "Start update"
+        # Load the SharePoint Assembly, using old style for backward compatibility with V1
+        [System.Reflection.Assembly]::LoadWithPartialName("Microsoft.SharePoint") | Out-Null
+
+        # Get a reference to the Administration WebService
+        $admService = [Microsoft.SharePoint.Administration.SPWebService]::ContentService
+
+        # Set the antivirus settings
+        if ($params.ContainsKey("AllowDownloadInfected")) { 
+            Write-Verbose -Message "Setting Allow Download"
```
BrianFarnhill: WIth this Write-Verbose, is this something that was around from your debugging when developing it? Just noticed you only write the output for the first property and not the others. Happy to have the logging in there for all so the verbose logs show which properties were actually hit but just wanted to be consistent across all the properties

Modules/xSharePoint/DSCResources/MSFT_xSPFarmAdministrators/MSFT_xSPFarmAdministrators.psm1
BrianFarnhill: Also its worth noting that this approach lines up with the [out of the box Group resource](https://msdn.microsoft.com/en-us/PowerShell/DSC/groupResource) so I'm happy to see us use the same approach

BrianFarnhill:
@ykuijs - What are your thoughts on changing the farm administrators one to a more generic group one so that it can be used to add a user to any specific group in any site collection? At a quick glance it appears the logic should work for that scenario, then we can include documentation in the resource (I'm working on shipping this at the moment as well) that details how to add users to the farm administrators group.

kilasuit:
From my POV the group resource could essentially have DOMAIN\Domain Users in there as a MembersToExclude and therefore force all accounts in the Domain out of the group which therefore is not a suitable use case IMO.

@BrianFarnhill Which Team/Repo looks after that resource from so that I can feed the above back to?

kilasuit:
I.e We have to equate for the people in all situations that may include this as a "best practice" in their resources and that really isn't a suitable scenario at all

BrianFarnhill:
@kilasuit Keep in mind though, we aren't resolving groups from AD in here, just adding the entry, so if I added domain users like you describe, that would just make there that "DOMAIN\Domain Users" wasn't in the SharePoint group here, but would leave the other entries in tact, so really it would stop you from adding "everyone" to the group essentially, not blocking us from adding others to the group at all. That's all based on the concept that there is no concept of a "Deny" permission in SharePoint.

kilasuit:
Ok so from a Config Mgmt POV you would have a V1 (Farm build) config determined - after that you are making V1.0.1 - Where you should only be amending the smallest possible aspect at a time - which in this case would be if you had in the Members Variable would perhaps be Domain\SPFarmAdmins,Domain\DomainAdmins (as it has a type of String) however if Managed via AD then the changes would infact be done in AD not in SP and therefore making the need for MembersToInclude & MembersToExclude completely irrelevant as the actual work is done in AD?

Even if we remove AD and think Claims then doesnt the same principle apply?

BrianFarnhill:
@kilasuit if you are managing the values in AD then you are correct, but not all customers choose to manage the membership of these groups in active directory and instead choose to manage them directly in SharePoint (especially if we talk about making this resource something for all groups instead of just the farm admins, as an AD group wont resolve the members when it comes time to assigning a task from workflow to a SharePoint group for example) so there are scenarios where the AD group thing won't really apply

kilasuit:
Even if not managed in AD then the String value for Members should be an array and there would still be no need for MembersToInclude and MembersToExclude

Either way it should accept either a Singular String for a single group - further resolved by AD or the accepted Identity management solution or should accept multiple entries.

And in this instance we are at Farm level Admin so it makes sense to make this as simplistic as possible without thinking about Item Level going forward for this resource in my opinion.

BrianFarnhill:
@kilasuit But back to the point made earlier, if you want to use it like that then you'll have a property for that, where you can define a specific list of users or groups from whatever your identity solution is and it will make sure that is what is applied, but if you want to manage it through a list of ensuring some are in and some are not and let exceptions to that be manually managed you can use the two other properties instead. So really we are providing two ways to do it based on however the administrator wants to manage it. I accept your point about wanting to keep it simple, but there are others who argue that the other two options is the better one for them as well - and since they aren't mutually exclusive and it lines up with how the group resource works we can do the same thing here.

Modules/xSharePoint/DSCResources/MSFT_xSPAntivirusSettings/MSFT_xSPAntivirusSettings.psm1
ykuijs: Correct, I used that for debugging. Unfortunately I found out that the Write-Verbose cmdlets inside the Invoke-Command scriptblock do not produce any output on the machine you are pushing the config from. So I would prefer to remove them in the scriptblocks.

ykuijs:
I agree with Brian on his last comment. From my POV it is all about providing flexibility in using the resource.
- If you want to enforce a specific list of users, great use the Members property. I think I would use this option the most.
- If you want to prevent a specific user or group from being added to the list, great use the MembersToExclude property.
- If you have the need for a specific user or group (for example a security department) to always be in the list, but would like to leave the rest up to the admins, great again: use the MembersToInclude property.

ykuijs:
@BrianFarnhill Let me look into the question about making it generic for groups. This might be an interesting once, since for the Farm Administrators you are specifically requesting the CentralAdmin web application. From a code perspective this is easy, but how to make the different in the DSC configuration easy to understand for the users of the resource. Any suggestions?

And I am thinking about making things clear: The xSPFarmAdministrators resource is for changing just one group, the Farm Admins and that is it. The name says it all.
We can then reuse the same logic/code for a new resource xSPGroup (or whatever we call it) where we can add/remove groups in site collections and maintain the members of it.

Modules/xSharePoint/DSCResources/MSFT_xSPAntivirusSettings/MSFT_xSPAntivirusSettings.psm1
BrianFarnhill: Are you sure? We've been seeing the verbose output of other logging from other resources turning up just fine?

ykuijs: Also when they are in the scriptblock that runs in the Invoke-Command process? I haven't seen any results of Write-Verbose when I use it in a Invoke-Command process. But I might be doing something wrong :-)

BrianFarnhill: Yep - we see the verbose output in the DSC extension in Azure for sure. I don't have an on-prem lab to test it with elsewhere but I've seen it using native commands like "Test-DSCConfiguration -Verbose" as well. Inside of the Invoke-xSharePointCommand statement we automatically bump the verbose output level up so we see everything. Makes for some large log files when you do something like set up a farm or a search topology for the first time, but we can drop custom messages out this way as well.

ykuijs: Just performed a small test. When running the script below, I do not see any verbose output:
$credential = Get-Credential
$server = "<server>"

$session = New-PSSession -ComputerName $server -Credential $Credential -Authentication CredSSP -ErrorAction Continue
$result = Invoke-Command -Session $session -ScriptBlock {Write-Verbose "This is a test"} -ArgumentList $test -Verbose
$result

Also found many articles on the Internet stating that the Verbose output in Invoke-Command will not work that well. Will test with the DSC resource as well.

BrianFarnhill: When we run things inside of our Invoke-xSharePointCommand though we get the output

    Invoke-xSharePointCommand -ScriptBlock { Write-Verbose "this is a test" } 

Now this may only be happening when we call our cmdlet without a credential (which is how it runs with PowerShell 5's PsDscRunAsCredential parameter instead of the InstallAccount we built in which will use a session like you tested with). Given the plan is to prefer the use of the PowerShell 5 way of running things, we can drop verbose output in to our script and see it that way.

ykuijs: Good one, will test that method as well. I have been testing the InstallAccount way until now

BrianFarnhill:
@ykuijs - I've finished my reivew, some notes on what is here:

 - When adding files to the project make sure you save the visual studio project as well and include it in the commits
 - All resources must have corresponding unit tests

I got on a roll when reviewing it and wrote these for you - you should see a [pull request in your repo](https://github.com/ykuijs/xSharePoint/pull/1) to pull from one of my branches in to your dev branch with these fixes - have a look over them and when you push them in to your dev branch they should turn up here automatically. After that I'll get this merged in to the main dev branch.

ykuijs:
@BrianFarnhill - I have been using Powershell ISE so far, instead of Visual Studio so I have not edited the Visual Studio project. Will check this out and switch over to VS.

Am not familiar with creating Unit tests. Will check your pull request on what you have done and make sure I include these in the future. Might be coming back to you with questions on this topic ;-)

BrianFarnhill:
@ykuijs No worries - email me questions as needed :)

BrianFarnhill:
@ykuijs Thanks for merging the changes in on this one, I'll merge and close this off now.

Issue #100, opened 2015-11-25
Prerequisite Installer Does Not Validate File Paths


HiltonGiesenow:
If you set the Prere installer to use Offline mode, and supply it with paths to the prereqs, it doesn't validate that they are actually available at the path supplied, and the installer gets stuck in an endless reboot loop.

NikCharlebois:
Thanks Hilton, actually, on Windows Server 2012 R2, even if the Prereq mode is set to online, I get stuck in an infinite loop. I will take a closer look at the event logs to see if I can figure out what the issue is.

BrianFarnhill:
So there was one issue we found with the parameters not being formatted correctly, the fix for which is in my current PR #97 along with a couple of other things, so that might be the cause of what you are looking for (the discussion on this is in issue number #88).

There is a larger issue here about the way things get caught in reboots here and us essentially relying on the prerequisite installer and how it behaves to meet our needs. I think ultimately we need to rewrite this resource to manually install the required prerequisites rather than calling the prerequisite installer. This will give us more control over the process and let us do more than run and exe and hope for the right exit code. What do you guys think?

HiltonGiesenow:
hmm, I'm a bit in two minds. On the one hand it would be much better to
have more control, but on the other we're reinventing the wheel. Perhaps
there's just a better way to get output from the prereq.exe. For instance,
when it failed for me, the gui showed up and gave me a link to the error
log - perhaps there's a way to check for an error log more explicitly? On
the other hand, that also would be ugly and we'd probably end up parsing
the log for the real error.

Perhaps it boils down to if we can reasonably and feasibly (time-wise)
rebuild this functionality and keep it up to date if/when it ever changes?
Ironically, what's the prereq installer is doing is pretty much what DSC
was designed to do - install windows features and msi's :-). If we really
wanted to, we could even look to download the installs to a known location
so that future (re)installs could naturally become offline installs - what
do you think? We should just do so in an idempotent way, like
"keepdownloads = $true" or something like that


-- Hilton
__________________________
Hilton Giesenow

(w) http://hilton.giesenow.com/
(m) +27 (0)83 642 6208

BrianFarnhill:
I'm closing this one now as the original issue about the prereq installer not validating the file paths was merged in through PR #101 - but I've also opened a new issue to discuss the approach of the prereq installer separately at #102

PR #101, opened 2015-11-25
Fix for 'Prerequisite Installer Does Not Validate File Paths' #100

HiltonGiesenow:
But this issue crops up in other ways - I think the fact that we just mark for a reboot for all cases when the process returns '0' should be looked at - we need to better differentiate issues.


BrianFarnhill:
@HiltonGiesenow Thanks for the contribution, we just need to do a couple of things to get this one through

 - Text formatting, we need to not use tabs, need to use spaces instead (there is a unit test failing for this)
 - I'm also seeing our other unit tests on this module failing because of the fact that it's throwing your exception instead of passing through, so we need to add some mocks and test for when the prereqs exist and when they don't

Ping me an email and I can talk you through these, or if needed I'll put it on my to do list to tidy up once I get my current PR pushed through.

BrianFarnhill:
@HiltonGiesenow Thanks for making the changes here, happy to merge this change in now

PR #107, opened 2015-12-01
Addition of xSPSessionStateService

wictorwilen:
Added xSPSessionStateService DSC Resource to configure the
SPSessionStateService

### DIFF
--- a/Modules/xSharePoint/DSCResources/MSFT_xSPSessionStateService/MSFT_xSPSessionStateService.psm1
+++ b/Modules/xSharePoint/DSCResources/MSFT_xSPSessionStateService/MSFT_xSPSessionStateService.psm1
@@ -1,0 +1,102 @@
+function Get-TargetResource
+{
+    [CmdletBinding()]
+    [OutputType([System.Collections.Hashtable])]
+    param
+    (
+        [parameter(Mandatory = $true)]  [System.String] $DatabaseName,
+        [parameter(Mandatory = $true)]  [System.String] $DatabaseServer,
+        [parameter(Mandatory = $true)]  [System.Boolean] $Enabled,
+        [parameter(Mandatory = $false)]  [System.UInt32] $SessionTimeout,
+        [parameter(Mandatory = $false)] [System.Management.Automation.PSCredential] $InstallAccount
+    )
+
+    Write-Verbose -Message "Getting SPSessionStateService info"
+    
+    $result = Invoke-xSharePointCommand -Credential $InstallAccount -Arguments $PSBoundParameters -ScriptBlock {
+        $params = $args[0]
+        
+        $svc = Get-SPSessionStateService
+        
+        return @{
+            DatabaseName = $svc.DatabaseId
+            DatabaseServer = $svc.DatabaseServer
+            Enabled = $svc.SessionStateEnabled
+            SessionTimeout = $svc.Timeout.TotalMinutes
+        }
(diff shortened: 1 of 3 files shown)


BrianFarnhill:
@wictorwilen Thanks for the script, looks good. We'll need some unit tests over this one as well though - I'm happy to write them real quick and send you a PR with the changes to merge in to this if needed though, otherwise happy for you to write them. Let me know.

wictorwilen:
Haven't written unit tests for PoSh previously, do you have any guidance?

BrianFarnhill:
@wictorwilen The tests we write are all done using Pester - if you have a look inside the Tests project you'll see all the stuff we have done previously. Basically we work by defining the different contexts in which this could run (e.g. "the state service isn't enabled", and "the state service is enabled") and then test that the get, test and set methods all run as they should in each condition. To simulate the conditions we mock the specific functions that would be called to return values represent each context we test against, and then we use Assert-MockCalled to determine that specific calls were made. 

Have a look at the tests for [xSPStateServiceApp](https://github.com/PowerShell/xSharePoint/blob/dev/Tests/xSharePoint/xSharePoint.xSPStateServiceApp.Tests.ps1) as a starting point as it's pretty simple but shouldn't be that different to what you are doing here. Also happy for you to ping me via email or skype if you want more detail or guidance.

wictorwilen:
Awesome, will take a look at it!

camiloborges:
Victor, I think checking the xSPStateServiceApp will make your life a lot easier indeed. 
I believe you just need to include a test "when there is a service and timeout is different".

Learning unit tests in PowerShell is really cool. :)

HiltonGiesenow:
Hey Wictor,

I'm pretty sure you'll figure it out quick, but I'm also happy to help if
you need - I spent some time just last week figuring it out, and we're on
roughly the same time zone whereas Brian might be asleep :-)


-- Hilton
__________________________
Hilton Giesenow

(w) http://hilton.giesenow.com/
(m) +27 (0)83 642 6208

BrianFarnhill:
@HiltonGiesenow Its cute the way you think that I use my nights to sleep instead of cutting code :P

HiltonGiesenow:
you know, I wasn't sure to mention that or not!


-- Hilton
__________________________
Hilton Giesenow

(w) http://hilton.giesenow.com/
(m) +27 (0)83 642 6208

BrianFarnhill:
@wictorwilen Sorry dude I didnt see that you added the tests, reviewing this one for you now.

BrianFarnhill:
@wictorwilen Scripts themselves look good so do the tests, just a couple of minor things that I'll fix after merging this one on because they are quick ones (but so you know for next time)

 * All files need to have an empty line at the end of them to help with the diff'ing GitHub does (I think I'll write a unit test for this and make it part of the test suite)
 * We've started adding documentation inside the MOF files that we use to generate the wiki pages for GitHub as well as the inbuilt help for inside PowerShell, have a look at [an example](https://github.com/PowerShell/xSharePoint/blob/dev/Modules/xSharePoint/DSCResources/MSFT_xSPCreateFarm/MSFT_xSPCreateFarm.schema.mof) to see what I mean, pretty straight forward but helps me make sure keep the doco up to date by keeping it in with the code like that.

Like I said though I'll touch them up after merging though because it'll take a few seconds to do. Thanks for the contribution!

PR #109, opened 2015-12-02
Added xSPVisioServiceApp

NikCharlebois:
Initial commit for xSPVisioServiceApp


BrianFarnhill:
@NikCharlebois Thanks for the PR, a couple of things I'll need before I can merge this one:

 - Need to fix the text formatting (look at the failing test, it sees tabs in MSFT_xSPVisioServiceApp.psm1 and MSFT_xSPVisioServiceApp.schema.mof)
 - Need a unit test set to cover this resource. Have a look at one of the simple service apps we have like the BCS one and pretty much copy that one

Let me know if you need any help with those so we can get this one merged in. Thanks Bud

PR #113, opened 2015-12-05
Fixes to prereq installer

BrianFarnhill:
Made changes to stop returning incorrect false results for SP2016, and remove the "looping" that is seen.


BrianFarnhill:
Closing as this has a bunch of the v0.8 merges that are still pending, will merge this separately later

PR #118, opened 2015-12-10
Added search topology resources

BrianFarnhill:


BrianFarnhill:
@caadam @camiloborges First one to code review wins - go!

camiloborges:
cool. I might be able to review it at some point today.

Modules/xSharePoint/DSCResources/MSFT_xSPSearchIndexPartition/MSFT_xSPSearchIndexPartition.psm1
```diff
@@ -0,0 +1,170 @@
...
+    )
+
+    $result = Invoke-xSharePointCommand -Credential $InstallAccount -Arguments $PSBoundParameters -ScriptBlock {
+        $params = $args[0]
+        $ConfirmPreference = 'None'
+
+        $ssa = Get-SPEnterpriseSearchServiceApplication -Identity $params.ServiceAppName      
+        $currentTopology = $ssa.ActiveTopology
+        
+        $IndexComponents = (Get-SPEnterpriseSearchComponent -SearchTopology $currentTopology | Where-Object { ($_.GetType().Name -eq "IndexComponent") -and ($_.IndexPartitionOrdinal -eq $params.Index) }).ServerName
```
camiloborges: hey. wouldn't this fail if you have Search App but no topology? Also, what happens if you don't have a Search App SVC that matches $params.ServiceAppName?

Tests/xSharePoint/xSharePoint.xSPSearchTopology.Tests.ps1
```diff
@@ -0,0 +1,307 @@
...
+                IndexPartition          = @($env:COMPUTERNAME)
+                FirstPartitionDirectory = "I:\SearchIndexes\0"
+            }
+
+            It "returns true from the test method" {
+                Test-TargetResource @testParams | Should Be $true
+            }
+        }
+    }
+}
```
camiloborges: Hey @BrianFarnhill 

Code seems to be good, rather long as expected, but seems ok. My only concern is, what if someone wanted to just 'ensure topology' from a supposedly existent Search Service App and that app doesn't exist, or if it exists and the topology is screwed to the point it  seems there is no topology.

Modules/xSharePoint/DSCResources/MSFT_xSPSearchIndexPartition/MSFT_xSPSearchIndexPartition.psm1
BrianFarnhill: Nope, I've used this in my labs to provision out a brand new service app, it works fine where there is no topology. Provisioned another one out from scratch last night.

Tests/xSharePoint/xSharePoint.xSPSearchTopology.Tests.ps1
BrianFarnhill: OK well if there is no topology it will add one, and i've tested it to update topologies that don't match in a number of component types and it appears to play nice making those changes. So from what I've seen in here it should be fine at enforcing an existing topology. But you are right, it won't fail gracefully if there is no search service app that it finds so I'll code that up and push it in here. Will let you know when it's done mate.

PR #120, opened 2015-12-12
Added web app policy resources

BrianFarnhill:


BrianFarnhill:
@camiloborges @caadam Another code review guys when you get a sec

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -0,0 +1,116 @@
...
+    Write-Verbose -Message "Getting cache accounts for $WebAppUrl"
+
+    $result = Invoke-xSharePointCommand -Credential $InstallAccount -Arguments $PSBoundParameters -ScriptBlock {
+        $params = $args[0]
+
+        $wa = Get-SPWebApplication -Identity $params.WebAppUrl -ErrorAction SilentlyContinue
+
+        if ($null -eq $wa) { return $null }
+        
+        if ($wa.Policies.UserName -contains $params.UserName) {
```
camiloborges: Hello. 
What if there is mixed auth and we have users with contoso\user and i:0#.w|contoso\chris ? If classic windows is added first then you won't be able to add the claims one.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -0,0 +1,116 @@
...
+            $policyObject = $wa.Policies | Where-Object { $_.UserName -eq $params.UserName }
```
camiloborges: shouldn't it be -ieq  to avoid negatives for contoso\user and CONTOSO\user?

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -0,0 +1,116 @@
...
+            }
+            "Full Control" {
+                $newRole = $wa.PolicyRoles.GetSpecialRole([Microsoft.SharePoint.Administration.SPPolicyRoleType]::FullControl)    
+            }
+            "Full Read" {
+                $newRole = $wa.PolicyRoles.GetSpecialRole([Microsoft.SharePoint.Administration.SPPolicyRoleType]::FullRead)    
+            }
+        }
+        
+        if ($wa.Policies.UserName -contains $params.UserName) {
```
camiloborges: same as before, classic x claims

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
BrianFarnhill: Nope, by default all comparison operators in PowerShell are case-insensitive. From https://technet.microsoft.com/en-us/library/hh847759.aspx - 

 > By default, all comparison operators are case-insensitive. To make a comparison operator case-sensitive, precede the operator name with a "c". For example, the case-sensitive version of "-eq" is "-ceq". To make the case-insensitivity explicit, precede the operator with an "i". For example, the explicitly case-insensitive version of "-eq" is "-ieq".

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
BrianFarnhill: Handled this in my last commit to also test for users that have been pumped through New-SPClaimsPrincipal to match the username from there as well

PR #122, opened 2015-12-15
Bug fix for offline install of SP2016 Beta2 prereqs

BrianFarnhill:


BrianFarnhill:
@caadam @camiloborges Quick code review on this bug fix if either of you has a couple of minutes, only very minor changes

camiloborges:
Hello. All good with the updates.
I think you'd have an easier life in the future if the requisites were a collection instead of named parameters. That will make it an interesting upgrade scenario if the pre-requisites are heavily changed in the future.

PR #123, opened 2015-12-16
Fixed incorrect name in schema

BrianFarnhill:


BrianFarnhill:
Minor typo fix in MOF schema that made it in to dev, correcting and pushing it right through

PR #126, opened 2015-12-19
Bug fix for property name

BrianFarnhill:


BrianFarnhill:
@ykuijs can you have a look at this seeing as you wrote the original resource? Just want to make sure the rename is OK to solve the issues I have been seeing with this one when testing

PR #139, opened 2015-12-23
xSPWorkManagement

camiloborges:
I can only hope we don't have the user profile sync conn files in this one, despite most commits been somehow related to it


camiloborges:
too much included in the PR, will start again

PR #141, opened 2015-12-24
xSPWordAutomationApp

camiloborges:
@ykuijs might have also got this one, but I am doign a PR anyway


BrianFarnhill:
@camiloborges Sorry bud, but the one that @ykuijs did in #156 is way more feature complete so we'll run with that one instead. Appreciate the effort though dude.

camiloborges:
No worries. I'd do the same :)


Camilo Borges
Mob:0488 056 964

PR #146, opened 2015-12-26
Bug fixes to MOF structure tests

BrianFarnhill:


BrianFarnhill:
@camiloborges @caadam Hey guys - code review when you get a minute. Figured out that some tests were showing false positives so fixed that up then patched up minor issues to do with it.

PR #159, opened 2016-01-24
X sp user profile property

camiloborges:


camiloborges:
I suspect I found a problem. I was reading the code and it seems it doesn't allow having a property without a mapped connection attribute

BrianFarnhill:
@camiloborges ping me with the details and we can chat about it

Modules/xSharePoint/DSCResources/MSFT_xSPUserProfileProperty/MSFT_xSPUserProfileProperty.schema.mof
```diff
@@ -0,0 +1,75 @@
...
+    UserOverridePrivacy = $false
+}
+
+*/
+
+
+[ClassVersion("1.0.0.0"), FriendlyName("xSPUserProfileProperty")]
+class MSFT_xSPUserProfileProperty : OMI_BaseResource
+{
+    [Key] string Name ;
```
TravisEz13: Adding descriptions would make this easier to use in ISE

Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1
```diff
@@ -32,6 +32,18 @@ function Get-xSharePointContentService() {
     return [Microsoft.SharePoint.Administration.SPWebService]::ContentService
 }
 
+
+function Get-xSharePointUserProfileSubTypeManager {
+    [CmdletBinding()]
+    param
+    (
+        $context
```
TravisEz13: Function parameter names should be Pascal case per [style guidance](https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md#parameter-names-should-use-pascalcase-)

Modules/xSharePoint/DSCResources/MSFT_xSPUserProfileProperty/MSFT_xSPUserProfileProperty.schema.mof
BrianFarnhill: @TravisEz13 We don't have descriptions on anything in xSharePoint yet - I'll add a backlog item to back capture this for all resources though

TravisEz13: That is sufficient.  Definitely not blocking.

Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1
TravisEz13: My main concern here is that we are internally consistent, at least in a module.  It makes it easier to read and review if we all use the same naming and casing standards.

camiloborges: Traviz. indeed, we are being inconsistent here :(
I will address this very one but I think we need a unit test to validate parameters are in the correct case. 
@BrianFarnhill  we might need to add this test in our 'Technical debt' backlog.

BrianFarnhill:
@camiloborges You need a pull from dev to make this mergable again

camiloborges:
@BrianFarnhill  go nuts

PR #177, opened 2016-02-10
Fixed bug with work management service app schema

BrianFarnhill:


BrianFarnhill:
@camiloborges Wanna have a quick look at this one mate?

camiloborges:
Looks good. Not sure we need a -ErrorAction when removing the service app, just because we will want to know if something fails

Camilo Borges
Mob:0488 056 964

BrianFarnhill:
@camiloborges I copied that from the changes in your previous pull request dude :P But you're right, I'll pull it out

PR #181, opened 2016-02-11
Version as optional parameter for the feature resource

wulfland:
There is opften the need to upgrade features by deactivating and activating them. This is handled with the feature version. 
Since the parameter is optional it is a non breaking change.

General: I thinks there are tons of farm solutions out there that won't be updated to the new app model. With the xSharePoint resource the solutions could be deployed to there own SP instances in the cloud. In this case the developers might need to reactivate features during deployment.


BrianFarnhill:
Happy with the logic about why it needs to be there, will review shortly and get back to you on the script.

Modules/xSharePoint/DSCResources/MSFT_xSPFeature/MSFT_xSPFeature.psm1
```diff
@@ -47,23 +49,55 @@ function Set-TargetResource
...
 
     $result = Invoke-xSharePointCommand -Credential $InstallAccount -Arguments $PSBoundParameters -ScriptBlock {
         $params = $args[0]
         
 
-        $runParams = @{}
-        $runParams.Add("Identity", $params.Name)
+        $runParams = @{ 
+            Identity = $params.Name 
+            Verbose = $false
```
BrianFarnhill: Can we take out the verbose = $false line. The Verbose logging is something we have aimed to use where it happens out of the box (so not specifying it false or true either way) so the output is consistent with what happens out of the box.

Modules/xSharePoint/DSCResources/MSFT_xSPFeature/MSFT_xSPFeature.psm1
```diff
@@ -47,23 +49,55 @@ function Set-TargetResource
...
+        }
+        
+        if ($params.FeatureScope -eq "Farm") 
+        {
+            $checkParams.Add($params.FeatureScope, $true)
+        } 
+        else
+        {
+            $checkParams.Add($params.FeatureScope, $params.Url)
+        }
```
BrianFarnhill: Rather than run this same logic from the get method, can you call Get-TargetResource to determine what you need here rather than duplicate the code?

Modules/xSharePoint/DSCResources/MSFT_xSPFeature/MSFT_xSPFeature.psm1
```diff
@@ -47,23 +49,55 @@ function Set-TargetResource
...
+        
+
         if ($params.Ensure -eq "Present") {
+            if (Get-SPFeature @checkParams -ErrorAction SilentlyContinue){
+                    
+                # Disable the feature first if it already exists.
+                $runParams.Add("Confirm", $false)    
+                Write-Verbose "Disable Feature '$($params.Name)' because it is already active at scope '$($params.FeatureScope)'..."
+                Disable-SPFeature @runParams
+            }
```
BrianFarnhill: I'm assuming this is the logic to disable the feature and reactivate where the version is different, yea? If so, rather than just assuming that it's there can we ensure that the version number is different here too (which if you use the Get-TargetResource method above you'll have easy access to the current values)

Modules/xSharePoint/DSCResources/MSFT_xSPFeature/MSFT_xSPFeature.psm1
```diff
@@ -80,12 +114,20 @@ function Test-TargetResource
...
     Write-Verbose -Message "Testing for feature $Name at $FeatureScope scope"
-    return Test-xSharePointSpecificParameters -CurrentValues $CurrentValues -DesiredValues $PSBoundParameters -ValuesToCheck @("Ensure")
+
+    $valuesToCheck = @("Ensure")
+
+    if ($Version) {
+        $valuesToCheck += @("Version")
+    }
+
+    return Test-xSharePointSpecificParameters -CurrentValues $CurrentValues -DesiredValues $PSBoundParameters -ValuesToCheck $valuesToCheck
```
BrianFarnhill: Can you pass version in here anyway to keep it clean - if version is not specified in PSBoundParameters then it will automatically be ignored by the Test-xSharePointSpecificParameters method, so just take the line as it was and add the "Version" bit to the existing array next to Ensure

README.md
```diff
@@ -99,6 +99,7 @@ Additional detailed documentation is included on the wiki on GitHub.
  * Removed Visual Studio project files, added VSCode PowerShell extensions launch file
  * Added xSPDatabaseAAG resource
  * Fixed bug with xSPWorkManagementServiceApp schema
+ * Added version as optional parameter for the feature resource to allow upgrading features to a specific version
```
BrianFarnhill: can you change "feature" to be "xSPFeteature" there so its clear the exact resource that it applies to:

    Added version as optional parameter for the xSPFeteature resource to allow upgrading features to a specific version

Tests/xSharePoint/xSharePoint.xSPFeature.Tests.ps1
```diff
@@ -146,5 +147,29 @@ Describe "xSPFeature" {
...
+
+            It "reactivates the feature in the set method" {
+                Set-TargetResource @testParams
+
+                Assert-MockCalled Disable-SPFeature -Times 1
+                Assert-MockCalled Enable-SPFeature -Times 1
+            }
+        }
     }
 }
```
BrianFarnhill: Can I get you to drop an empty line at the end of this file as well

Modules/xSharePoint/DSCResources/MSFT_xSPFeature/MSFT_xSPFeature.psm1
wulfland: This results in another roundtrip to the Server and loading SharePoint snapin again. I took it out for performance reasons - but I understand your concern for maintainability and moved it back in.
But I wouldn’t check the version here. If we’re at the point, that we want to enable a feature at a given scope and it is already activated we have to deactivate it first anyway. Ok with that? I passed in all current values anyway to easily add more logic to it.

PR #187, opened 2016-02-18
Fixed bug with UPS Connection

BrianFarnhill:
@camiloborges Review for you mate after we talked about this one already


camiloborges:
@BrianFarnhill you looking at doing a release this weekend ?

Camilo Borges

PR #188, opened 2016-02-18
Fixed issue with kerberos and anon auth

BrianFarnhill:
Quickie code review @camiloborges @anlynes @caadam @ykuijs


ykuijs:
Looks good!

PR #191, opened 2016-02-20
Updated examples to include install resources

BrianFarnhill:


Modules/xSharePoint/Examples/Single Server/SharePoint.ps1
```diff
@@ -41,6 +41,35 @@ Configuration SharePointServer
...
+        }
+        xSPInstall InstallSharePoint {
+            Ensure = "Present"
+            BinaryDir = $ConfigurationData.NonNodeData.SharePoint.Binaries.Path
+            ProductKey = $ConfigurationData.NonNodeData.SharePoint.ProductKey
+            DependsOn = "[xSPInstallPrereqs]InstallPrereqs"
+        }
 
         #**********************************************************
         # Basic farm configuration
```
ykuijs: The CreateFarm resource should depend on InstallSharePoint, just like with the Small Farm example. Dependancy is now missing completely.

ykuijs:
Small addition, the rest looks great.

PR #196, opened 2016-02-26
Added support for Server 2016 TP

BrianFarnhill:
@camiloborges @ykuijs @anlynes @caadam Code review boys - this one is a quickie, just had to modify the windows features to check on the prereq installer to get 2016 to pass the prereq test on server 2016 TP4. For testing purposes, this was using the RC build of 2016


camiloborges:
All good. 'Massive change'

Camilo Borges

PR #209, opened 2016-03-09
Search hybrid

NikCharlebois:
Added params checks for Hybrid Search...Unfortunately, I had not created a unique branch for my Reverse DSC change, so I am carrying these along.... I promise I'll get the hang of GitHub....some day


BrianFarnhill:
@NikCharlebois I can't pull the reverse changes in yet until we get through the stuff we discussed around that and test it a bit more. Are you able to put the cloud SSA stuff in its own branch? If not I can lift the code from here and put it in a new one if that helps?

camiloborges:
You might need to branch it from the one you have at the moment to remove the reverse.
Welcome to the club. Took me so long to get the hang of it. Nowadays I run git console for most things. And I couldn't f'ing get any PR or Pull done right.

Camilo Borges

PR #213, opened 2016-03-10
Bug fix for SPServer objects that use FQDNs

BrianFarnhill:
We initially saw this bug about 10 months ago but couldn't reproduce so left it alone. It's finally reared its head again and I've been able to squash it. The scenario is that in some SharePoint installations you will see SPServer objects referred to using a FQDN (like "server.domain.name") instead of just the server name. This breaks xSPServiceInstance, xSPUserProfileSync, xSPSearchTopology and xSPDistributedCacheService as they all refer to these server objects in some form, so this PR covers a fix for each that sees us try to load a service instance from the computer name alone first, and if that returns null we re-run the check using the FQDN. 

Code review time @caadam @camiloborges @ykuijs @anlynes

### DIFF
--- a/Modules/xSharePoint/DSCResources/MSFT_xSPDistributedCacheService/MSFT_xSPDistributedCacheService.psm1
+++ b/Modules/xSharePoint/DSCResources/MSFT_xSPDistributedCacheService/MSFT_xSPDistributedCacheService.psm1
@@ -172,7 +172,10 @@ function Set-TargetResource
                 $domain = (Get-CimInstance -ClassName Win32_ComputerSystem).Domain
                 $currentServer = "$currentServer.$domain"
                 $serviceInstance = Get-SPServiceInstance | Where-Object { ($_.Service.Tostring()) -eq "SPDistributedCacheService Name=AppFabricCachingService" -and ($_.Server.Name) -eq $currentServer }
-            }                        
+            }
+            if ($serviceInstance -eq $null) {
+                throw "Unable to locate a distributed cache service instance on $($env:computername) to remove"
+            }               
             $serviceInstance.Delete() 
             
             Remove-SPDistributedCacheServiceInstance
--- a/Modules/xSharePoint/DSCResources/MSFT_xSPDistributedCacheService/MSFT_xSPDistributedCacheService.psm1
+++ b/Modules/xSharePoint/DSCResources/MSFT_xSPDistributedCacheService/MSFT_xSPDistributedCacheService.psm1
@@ -103,6 +103,13 @@ function Set-TargetResource
                         $count = 0
                         $maxCount = 30
 
+                        # Attempt to see if we can find the service with just the computer name, or if we need to use the FQDN
+                        $si = Get-SPServiceInstance -Server $currentServer | Where-Object { $_.TypeName -eq "Distributed Cache" }                    
+                        if ($null -eq $si) { 
+                            $domain = (Get-CimInstance -ClassName Win32_ComputerSystem).Domain
+                            $currentServer = "$currentServer.$domain"
+                        }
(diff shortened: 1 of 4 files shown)


Modules/xSharePoint/DSCResources/MSFT_xSPSearchTopology/MSFT_xSPSearchTopology.psm1
```diff
@@ -83,7 +82,14 @@ function Set-TargetResource
...
         foreach($searchServer in $AllSearchServers) {
-            $searchService = Get-SPEnterpriseSearchServiceInstance -Identity $searchServer
+            
+            $searchService = Get-SPEnterpriseSearchServiceInstance -Identity $searchServer -ErrorAction SilentlyContinue
+            if ($searchService -eq $null) {
+                $domain = (Get-CimInstance -ClassName Win32_ComputerSystem).Domain
+                $searchServer = "$searchServer.$domain"
+                $searchService = Get-SPEnterpriseSearchServiceInstance -Identity $searchServer    
+            }
+
```
ykuijs: What happens if after this change the result is still $null?? As far as I can see there is no additional check.

Modules/xSharePoint/DSCResources/MSFT_xSPDistributedCacheService/MSFT_xSPDistributedCacheService.psm1
```diff
@@ -161,6 +168,11 @@ function Set-TargetResource
         Write-Verbose -Message "Removing distributed cache to the server"
         Invoke-xSharePointCommand -Credential $InstallAccount -ScriptBlock {
             $serviceInstance = Get-SPServiceInstance | Where-Object { ($_.Service.Tostring()) -eq "SPDistributedCacheService Name=AppFabricCachingService" -and ($_.Server.Name) -eq $env:computername }
+            if ($null -eq $serviceInstance) { 
+                $domain = (Get-CimInstance -ClassName Win32_ComputerSystem).Domain
+                $currentServer = "$currentServer.$domain"
+                $serviceInstance = Get-SPServiceInstance | Where-Object { ($_.Service.Tostring()) -eq "SPDistributedCacheService Name=AppFabricCachingService" -and ($_.Server.Name) -eq $currentServer }
+            }
```
ykuijs: What happens if after this change the result is still $null?? As far as I can see there is no additional check.

Modules/xSharePoint/DSCResources/MSFT_xSPUserProfileSyncService/MSFT_xSPUserProfileSyncService.psm1
```diff
@@ -84,8 +89,13 @@ function Set-TargetResource
...
         
-
-        $syncService = Get-SPServiceInstance -Server $env:COMPUTERNAME | Where-Object { $_.TypeName -eq "User Profile Synchronization Service" }
+        $currentServer = $env:COMPUTERNAME
+        $syncService = Get-SPServiceInstance -Server $currentServer | Where-Object { $_.TypeName -eq "User Profile Synchronization Service" }
+        if ($null -eq $syncService) { 
+            $domain = (Get-CimInstance -ClassName Win32_ComputerSystem).Domain
+            $currentServer = "$currentServer.$domain"
+            $syncService = Get-SPServiceInstance -Server $currentServer | Where-Object { $_.TypeName -eq "User Profile Synchronization Service" }
+        }
```
ykuijs: What happens if after this change the result is still $null?? As far as I can see there is no additional check.

ykuijs:
Added the same question on several places in the code. The rest looks fine.

PR #216, opened 2016-03-16
Fixed case sensitive check of farm admins resource

BrianFarnhill:
@ykuijs Can I get you to review this one dude since it was your resource to begin with? The issue is that I was seeing SharePoint return "domain\user" where as my config had "Domain\User" (casing was the only difference) and it would fail a test, despite the user being in the group already. Just switched the Contains() method out for the PowerShell contains operator which isn't case sensitive.


ykuijs:
Looks perfect!

PR #222, opened 2016-03-17
Fixed bug with default URLs being changed in xSPAlternateUrl

BrianFarnhill:
@caadam @camiloborges @anlynes @ykuijs Code review time - This is a quick one to account for the AAMs being changed in the default zone.


ykuijs:
Looks good!

Issue #148, opened 2015-12-28
$Ensure Parameter


camiloborges:
Hey.
I noticed some of the resources have the 'Ensure' parameter. Shouldn't all of them have it?

BrianFarnhill:
Pretty much yes - feel free to call it out in the comments on any outstanding issues that don't have it, and we can raise issues for ones that are missing it.

camiloborges:
Cool. deal

ykuijs:
I would suggest only to include them when it makes sense. For example
- The xSPFarmAdministrators resource doesn't need to have the ensure parameter, since the currently specified parameters are (imho) sufficient.
- A Service Application resource requires one, since you need to be able to specify it the SA needs to be present or not.

camiloborges:
agree with @ykuijs , not all need it, but many do. I went through the list we have in the latest release and documented them.

resources with missing Ensure
MSFT_xSPWebAppPolicy: Ensure would remove given user from web app policies
MSFT_xSPWebApplication 
MSFT_xSPVisioServiceApp
MSFT_xSPUsageApplication
MSFT_xSPSubscriptionSettingsServiceApp
MSFT_xSPStateServiceApp
MSFT_xSPSite
MSFT_xSPSessionStateService
MSFT_xSPServiceAppPool
MSFT_xSPSecureStoreServiceApp
MSFT_xSPSearchServiceApp
MSFT_xSPManagedPath
MSFT_xSPManagedMetadataServiceApp
MSFT_xSPManagedAccount
MSFT_xSPBCSServiceApp
MSFT_xSPAppManagementServiceApp

BrianFarnhill:
Pretty happy with this list - I'll put this in to the enhancement category and we can capture this in a future udpate.

camiloborges:
Hey, I think we need to address this for V1.0. It believe we'll also benefit from having it when we get the reverse module sorted out

BrianFarnhill:
I'm taking this one to get it in for 1.0

BrianFarnhill:
Merged in to dev, closing this issue

PR #227, opened 2016-03-21
Added ensure properties throughout

BrianFarnhill:
@camiloborges @ykuijs @caadam @anlynes - Monster code review here guys. As per #148 I've gone through and:

* Added an ensure property to all resources that make sense to have one
* Checked that the ensure property is optional on all resources that have it, and that it defaults to "present"
* Updates tests to line up with the new approach

Its a big one, but important ahead of the 1.0 release. Let me know who wants to start the code review


ykuijs:
@BrianFarnhill Unit tests are failing.

Will review this PR tomorrow. Will have time available then.

BrianFarnhill:
@ykuijs Fixed the tests - I screwed something up in my merge from dev. Fixed in the updated merge though, so good to review when ready

Modules/xSharePoint/DSCResources/MSFT_xSPAppManagementServiceApp/MSFT_xSPAppManagementServiceApp.psm1
```diff
@@ -94,14 +117,16 @@ function Test-TargetResource
...
     )
     
+    $PSBoundParameters.Ensure = $Ensure
+    
     Write-Verbose -Message "Testing for App management Service Application '$Name'"
     $CurrentValues = Get-TargetResource @PSBoundParameters
     
-    if ($null -eq $CurrentValues) { return $false }
-    return Test-xSharePointSpecificParameters -CurrentValues $CurrentValues -DesiredValues $PSBoundParameters -ValuesToCheck @("ApplicationPool")
+    return Test-xSharePointSpecificParameters -CurrentValues $CurrentValues -DesiredValues $PSBoundParameters -ValuesToCheck @("ApplicationPool", "Ensure")
```
ykuijs: Should we test for DatabaseName and DatabaseServer as well??

Modules/xSharePoint/DSCResources/MSFT_xSPUsageApplication/MSFT_xSPUsageApplication.psm1
```diff
@@ -129,8 +148,12 @@ function Test-TargetResource
 
     $CurrentValues = Get-TargetResource @PSBoundParameters
     Write-Verbose -Message "Testing for usage application '$Name'"
-    if ($null -eq $CurrentValues) { return $false }
-    return Test-xSharePointSpecificParameters -CurrentValues $CurrentValues -DesiredValues $PSBoundParameters -ValuesToCheck @("UsageLogCutTime", "UsageLogLocation", "UsageLogMaxFileSizeKB", "UsageLogMaxSpaceGB")
+    $PSBoundParameters.Ensure = $Ensure
+    if ($Ensure -eq "Present") {
+        return Test-xSharePointSpecificParameters -CurrentValues $CurrentValues -DesiredValues $PSBoundParameters -ValuesToCheck @("UsageLogCutTime", "UsageLogLocation", "UsageLogMaxFileSizeKB", "UsageLogMaxSpaceGB")
```
ykuijs: Shouldn't the ValuesToCheck also include Ensure??

Modules/xSharePoint/DSCResources/MSFT_xSPWebApplication/MSFT_xSPWebApplication.psm1
```diff
@@ -125,17 +151,17 @@ function Test-TargetResource
...
 
     $CurrentValues = Get-TargetResource @PSBoundParameters
     Write-Verbose -Message "Testing for web application '$Name'"
-    if ($null -eq $CurrentValues) { return $false }
-
+    $PSBoundParameters.Ensure = $Ensure
     $testReturn = Test-xSharePointSpecificParameters -CurrentValues $CurrentValues `
                                                      -DesiredValues $PSBoundParameters `
-                                                     -ValuesToCheck @("ApplicationPool")
+                                                     -ValuesToCheck @("Ensure")
```
ykuijs: Why are you only checking for the Ensure parameter and not the others?

ykuijs:
Man this was a big one :-) Only a few comments, the rest looks good.

Modules/xSharePoint/DSCResources/MSFT_xSPWebApplication/MSFT_xSPWebApplication.psm1
BrianFarnhill: Because it's the only thing we will change if it's not set correctly. When we add logic to the set method later to update other properties we can put them back in, but at this point it's the only thing that actually makes a difference if we want to call the set method.

Modules/xSharePoint/DSCResources/MSFT_xSPAppManagementServiceApp/MSFT_xSPAppManagementServiceApp.psm1
BrianFarnhill: No as we don't currently change those in the set method, so there is no point in letting them return false from the test method. Ensure will let us know if it needs to be provisioned, and ApplicationPool is the only other one that acts in the set method.

PR #231, opened 2016-03-22
Added basic CT Hub URL parameter

BrianFarnhill:
@ykuijs @caadam @camiloborges @anlynes - Code review. Quick one here, just added an option to provision MMS with a content type hub URL. I'll add a work item tagged as technical debt to track the fact that we don't monitor this URL and update it once the MMS is provisioned (found an easy way to set it, but no easy way to retrieve the current value) so we can revisit this later, but I wanted to get the basic property in so we didn't need to make a schema change later to add the property.

### DIFF
--- a/Modules/xSharePoint/DSCResources/MSFT_xSPManagedMetadataServiceApp/MSFT_xSPManagedMetaDataServiceApp.psm1
+++ b/Modules/xSharePoint/DSCResources/MSFT_xSPManagedMetadataServiceApp/MSFT_xSPManagedMetaDataServiceApp.psm1
@@ -8,6 +8,7 @@ function Get-TargetResource
         [parameter(Mandatory = $true)]  [System.String] $ApplicationPool,
         [parameter(Mandatory = $false)] [System.String] $DatabaseServer,
         [parameter(Mandatory = $false)] [System.String] $DatabaseName,
+        [parameter(Mandatory = $false)] [System.String] $ContentTypeHubUrl,
         [parameter(Mandatory = $false)] [System.Management.Automation.PSCredential] $InstallAccount      
     )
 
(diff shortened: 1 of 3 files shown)


Modules/xSharePoint/DSCResources/MSFT_xSPManagedMetadataServiceApp/MSFT_xSPManagedMetaDataServiceApp.psm1
```diff
@@ -62,6 +64,10 @@ function Set-TargetResource
             
 
             if ($params.ContainsKey("InstallAccount")) { $params.Remove("InstallAccount") | Out-Null }
+            if ($params.ContainsKey("ContentTypeHubUrl")) {
```
camiloborges: Don't we have a method inside Util to achieve this? 
That is the only comment on this change. :)

BrianFarnhill: I think we do, but there are a heap of places that would need to switch over to this, so we could pick this up as part of a wider clean up exercise. So I'll leave this one for the time being and investigate that cleanup.

Modules/xSharePoint/DSCResources/MSFT_xSPManagedMetadataServiceApp/MSFT_xSPManagedMetaDataServiceApp.psm1
```diff
@@ -62,6 +64,10 @@ function Set-TargetResource
...
+                $params.Add("HubUri", $params.ContentTypeHubUrl)
```
camiloborges: Sorry, I lied I have another comment. We need a note added somewhere highlighting this value isn't currently updated after the initial provisioning

BrianFarnhill: You mean like I did in the MOF description field which goes in to the PowerShell help and the wiki documentation?

camiloborges: exactly, my bad. I can't read shit in Canberran English.

BrianFarnhill: lol, thanks bud - I'll merge this in now

Issue #147, opened 2015-12-27
Extend xSPWebAppPolicy resource


ykuijs:
The current version of the xSPWebAppPolicy resource is only able to add new users. It is not able to remove users (Ensure=Absent) or enforce a specific list of users. See xSPFarmAdministrator group as an example.

Also: I haven't been able to test this yet, but reviewing the code I think the code does not handle claims users and classic users the correct way. The resource needs to be able to make a distinction between these two types of users. Using Central Admin you are only able to add claims users, but by using the SPWebApplication.GrantAccessToProcessIdentity method you can add classic identities. This is required in some cases, like using remote PowerShell to access specific properties (e.g. AllowDesigner) of the SPSite object.

BrianFarnhill:
@ykuijs We did test the claims stuff in that it assumes if it can't find a match for the user through the basic string name then it tries to interpret it as a claims identity. So the theory is that if you specify it through the DSC you only user user the non-claims version of it and it'll get matched either way. If you test it and see that it does something different let me know though. Happy to agree to the other changes you opened with though

ykuijs:
Am going to test this one and implement the changes suggested earlier.

ykuijs:
@BrianFarnhill : Ok, have tested it. Here are the results:
As you specified, when the policy contains the claim value for the user and the resource specifies the domain\user format user, the resource returns true. This means that there is no way anymore (using DSC) you can add the domain\user format. However as specified in my first post, in some cases this is required to have SharePoint work properly.

I therefore would like to suggest to change this to match the string in the resource with the string in the policy. That way, you have to specify the claims value if you would like to add the claims value and the domain\user value if your want to add the domain\user value.

ykuijs:
Working on this one

ykuijs:
Also including #144

BrianFarnhill:
@ykuijs Ping me to talk about this one again - I can't remember the spcifics of how we were approaching this and want to be sure we don't lose functionality. Thanks dude

ykuijs:
Check your mail

ykuijs:
New addition after a discussion with Brian:
Add SetCacheAccountsPolicy (boolean) property to make sure the Cache Accounts are also added (in claims format) to the policy.

BrianFarnhill:
This has been merged to dev, closing this issue

PR #233, opened 2016-03-22
Updated xSPWebAppPolicy resource

ykuijs:
As described in issue #147 and after a lot of issues with Pester, this update is finally complete!

It also contains the SetCacheAccountsPolicy parameter as discussed with @BrianFarnhill over the mail.

Code review time!!

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/xsharepoint/233)


BrianFarnhill:
@ykuijs Unit tests are failing - looks like there are spaces instead of tabs in the file there dude.

ykuijs:
Already fixed that yesterday, but looks like I did something wrong. Attempt 2 has succeeded :-)

BrianFarnhill:
@ykuijs Starting review now - looks like changes to dev are causing a merge conflict, can you have a look at that?

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -19,32 +44,32 @@ function Get-TargetResource
...
-            }
-            return $null 
         }
+
+        if ($wa.Properties.ContainsKey("portalsuperuseraccount")) {
+            $returnval.SuperUserAccount = $wa.Properties["portalsuperuseraccount"]
+        }
+        if ($wa.Properties.ContainsKey("portalsuperreaderaccount")) {
+            $returnval.SuperReaderAccount = $wa.Properties["portalsuperreaderaccount"]
+        }
```
BrianFarnhill: SuperUserAccount and SuperReaderAccount aren't parameters on the resource - SetCacheAccountsPolicy is. We need to make sure that the get method returns the same parameters as what goes in to it. So what we should be doing here is returning a true/false about whether or not the cache accounts are currently set or not.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -56,63 +81,260 @@ function Set-TargetResource
...
+    }
+
+    foreach ($member in $MembersToInclude) {
+        if (($member.ActAsSystemAccount -eq $true) -and ($member.PermissionLevel -ne "Full Control")) {
+            throw "MembersToInclude Parameter: You cannot specify ActAsSystemAccount with any other permission than Full Control"        
+        }
+    }
+    
+    $result = Invoke-xSharePointCommand -Credential $InstallAccount -Arguments @($PSBoundParameters,$PSScriptRoot) -ScriptBlock {
```
BrianFarnhill: $result is unused here (I'm cleaning this up as I go through some other resources as well) so can we take that off the front of this line?

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -56,63 +81,260 @@ function Set-TargetResource
...
-                if (($null -ne $claimsPrincipal) -and ($claimsPrincipal.Value -eq $params.UserName)) {
-                    $policyObject = $wa.Policies | Where-Object { $_.UserName -eq $userName }
+        $members = @()
+        foreach ($policy in $wa.Policies) {
+            $member = @{}
+            $member.Username = $policy.UserName
+            $member.PermissionLevel = $policy.PolicyRoleBindings.Name
+            $member.ActAsSystemAccount = $policy.IsSystemUser
+            $members += $member
+        }
```
BrianFarnhill: This looks like a double up of whats in the get method - can we just call the get method and pass that in here rather than code it up twice? There are a few examples of where we are doing this (I think you would have reviewed one or two of them in my stuff recently to know what I'm talking about)

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -56,63 +81,260 @@ function Set-TargetResource
...
+                    PermissionLevel = "Full Read"
+                }
+                $allMembers += $psuAccount
+                
+                $psrAccount = @{
+                    UserName = $wa.Properties["portalsuperreaderaccount"]
+                    PermissionLevel = "Full Read"
+                }
+                $allMembers += $psrAccount
+            }
```
BrianFarnhill: If the web app uses claims then we need to add the claims user as well as the standard user. Have a look at [my changes](https://github.com/PowerShell/xSharePoint/pull/234/files#diff-640dbcab1e86d47f40c29e5f97bfa399R100) in #234 to see how I'm doing this there

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -56,63 +81,260 @@ function Set-TargetResource
...
+                    UserName = $wa.Properties["portalsuperuseraccount"]
+                    PermissionLevel = "Full Read"
+                }
+                $allMembers += $psuAccount
+                
+                $psrAccount = @{
+                    UserName = $wa.Properties["portalsuperreaderaccount"]
+                    PermissionLevel = "Full Read"
+                }
+                $allMembers += $psrAccount
```
BrianFarnhill: Same remark here about the cache accounts need claims and default modes added. Can we consolidate this so we don't have this if statement in the script twice though? Not fussed about leaving it here, just noticed the repetition and wanted to call it out.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -56,63 +81,260 @@ function Set-TargetResource
...
+                            "Full Read" {
+                                $newPolicy.PolicyRoleBindings.Add($fullRead)
+                            }
+                        }
+                    }
+                    if ($member.ActAsSystemAccount) {
+                        $newPolicy.IsSystemUser = $member.ActAsSystemAccount
+                    }
+                }
+                $wa.Update()
```
BrianFarnhill: Did you see issues calling Update() multiple times on the web app? You should be able to call it just once at the end of adding all the accounts and making changes rather than once at the bottom of the foreach loop there

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -56,63 +81,260 @@ function Set-TargetResource
...
 
-            $wa.Update()
+                $policy = $wa.Policies | Where-Object { $_.UserName -eq $member.UserName }
+
+                if ($policy -ne $null) {
+                    # User exists. Delete user
+                    Write-Verbose -Verbose "User $($member.UserName) exists, deleting"
+                    Remove-WebAppPolicy $wa.Policies $member.UserName
+                }
+                $wa.Update()
```
BrianFarnhill: Same remark here about calling update multiple times - worth checking out as well

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -56,63 +81,260 @@ function Set-TargetResource
...
+        if ($params.MembersToExclude) {
+            Write-Verbose -Verbose "Processing MembersToExclude parameter"
+
+            $psuAccount = $wa.Properties["portalsuperuseraccount"]
+            $psrAccount = $wa.Properties["portalsuperreaderaccount"]
+
+            Import-Module (Join-Path $ScriptRoot "..\..\Modules\xSharePoint.WebAppPolicy\xSPWebAppPolicy.psm1" -Resolve)
+            
+            foreach ($member in $params.MembersToExclude) {
+                if (($psuAccount -eq $member.UserName) -or ($psrAccount -eq $member.UserName)) {
```
BrianFarnhill: You need to check for the claims versions of the usernames as well as the default DOMAIN\user names here

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -125,16 +347,108 @@ function Test-TargetResource
...
+        [parameter(Mandatory = $false)] [System.Boolean] $SetCacheAccountsPolicy,
         [parameter(Mandatory = $false)] [System.Management.Automation.PSCredential] $InstallAccount
     )
 
     $CurrentValues = Get-TargetResource @PSBoundParameters
     Write-Verbose -Message "Testing web app policy for $UserName at $WebAppUrl"
     if ($null -eq $CurrentValues) { return $false }
-    return Test-xSharePointSpecificParameters -CurrentValues $CurrentValues -DesiredValues $PSBoundParameters -ValuesToCheck @("PermissionLevel", "ActAsSystemUser")
+
+    if (($SetCacheAccountsPolicy) -and (-not($CurrentValues.ContainsKey("SuperUserAccount")) -or -not($CurrentValues.ContainsKey("SuperReaderAccount")))) {
```
BrianFarnhill: So this goes to the comment in the get method. It should be returning a true/false about whether the cache values are configured. In the test method then you can just check that the true/false matches (if they want to do the cache accounts) and the problem about not having the properties set should be an issue for the set method - we should be able to return that they aren't configured without an exception in the get/test methods, in theory the only place that we should see an exception is the set method as thats the place that is becomes an issue for us. Does that make sense?

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
```diff
@@ -125,16 +347,108 @@ function Test-TargetResource
...
+                UserName = $CurrentValues.SuperUserAccount
+                PermissionLevel = "Full Read"
+            }
+            $allMembers += $psuAccount
+            
+            $psrAccount = @{
+                UserName = $CurrentValues.SuperReaderAccount
+                PermissionLevel = "Full Read"
+            }
+            $allMembers += $psrAccount
```
BrianFarnhill: Ditto for the comment about claims accounts here as well

Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1
```diff
@@ -310,4 +310,15 @@ function Set-xSharePointObjectPropertyIfValueExists() {
     }
 }
 
+# Due to issues with unit testing, this function has been created so we are able to mock this function.
+function Remove-WebAppPolicy() {
```
BrianFarnhill: Why didn't this end up in your web app policy module?

Modules/xSharePoint/Modules/xSharePoint.WebAppPolicy/xSPWebAppPolicy.psm1
```diff
@@ -0,0 +1,99 @@
...
+function GetUserFromCollection() {
+    Param (
+        [Parameter(Mandatory=$true)] 
+        [Array] $collection,
+        [Parameter(Mandatory=$true)] 
+        [String] $user
+    )
+
+    foreach ($item in $collection) {
+        if ($item.Username.ToLower() -eq $user.ToLower()) { return $item }
```
BrianFarnhill: You shouldn't need to do the ToLower() on these, powershell will do a case-insensitive compare with "-eq"

xSharePoint.sln
```diff
@@ -0,0 +1,39 @@
+﻿
```
BrianFarnhill: Can you remove the SLN file here? We have moved on from using Visual Studio 2015 to Visual Studio code

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: Why is this necessary? Is this a requirement from DSC or something we specified?

The reason I did this is that I need these values in the Test-TargetResource and wanted to prevent another call to Invoke-xSharePointCommand in that method. Since I was already requesting information from the Get method, I included these parameters into the results set.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: Will do!

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: Correct, this is double. The reason was what we discussed via mail: I wanted to prevent another call to Invoke-xSharePointCommand to prevent the additional overhead.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: True, realized that when I reviewed your changes to the Cache Account resource. Will change this.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: No, didn't see any issues. Will look into this

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: True, like the other comment: Realized this when I reviewed your changes to the Cache Account resource. Will change this.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: I see what you are getting at: The Get method should perform a test if the cache accounts are included in the policy, what I am now doing in the Test method.

The issue I have here is that we have a combination of datasets: The user specifies Members (or MembersToInclude) and the cache accounts come from SharePoint. These two datasets need to be checked against the Members in the WebApp policy, which is just one dataset. So if I return True in the Get method, I still need the account names in the Test method to generate the combined result set. That is why I also return the two accounts, so I can construct the result set and compare that with the Members from the WebApp policy.

Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1
ykuijs: This was the only place I was able to Mock the function in the Pester test.

Modules/xSharePoint/Modules/xSharePoint.WebAppPolicy/xSPWebAppPolicy.psm1
ykuijs: Will change it.

xSharePoint.sln
ykuijs: Ok, no clue how it got in here. I am also using VS Code. Will remove it.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: According to the following article, the account should be entered in the WebApp properties in claims or classic format and then equally in the WebApp policy:
https://technet.microsoft.com/en-us/library/ff758656.aspx
First Step 14 in To create the user accounts by using Central Administration
Then Step 3 in To add the user accounts to the web application by using Windows PowerShell

So when the web app is in Claims mode, the account in the Cache Accounts should be entered in Claims format and this resource will automatically pick up that format and configure it accordingly. So I think the Cache Accounts resource should be changed and this resource is fine.

BrianFarnhill: OK so basically if the web app is in claims mode we add them in claims format to web app policy and the property bag, if not then in default format. I'll make the updates to the logic here and push them back in when ready.

Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1
BrianFarnhill: Really? You should be able to put it in to the util module and then when you mock specify the module name:

    Mock Remove-WebAppPolicy { } -ModuleName "xSharePoint.Util"

If you can't get it let me know and I'll take a look.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
BrianFarnhill: Having a get method that doesn't match the schema will cause some errors in PowerShell though - so if you need them in the test method you might need to retrieve them again there instead.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
BrianFarnhill: As per email discussion, this won't cause an issue going forward with PowerShell 5 so we should be fine to reuse the logic in the get method.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
BrianFarnhill: Yea this is a DSC requirement - the get method needs to match the schema, Powershell 5 will start throwing warnings and errors if it sees output that doesn't match the schema, so you might need to retrieve these in the test method separately if you need them.

ykuijs: Ok, will check this out. I am using PS5, but haven't seen any errors yet when I changed the output. Only when the input schema does not match the schema in the schema.mof file, like missing or differently configured parameters.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: Ok, will change this

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: Ok, I will leave this resource as is. I will however test a claims formatted user in the property bag. Haven't done that yet.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
ykuijs: As mentioned earlier, haven't seen any issues yet. But will update the resource anyways.

Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1
ykuijs: Have tried this, but the only location where I got it working was in the generic util file. Will do some more testing and let you know.

Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1
BrianFarnhill: Even if it's not throwing errors, it's definitely meant to - I can't remember the specific place I saw errors but I think it was when using Get-DscConfiguration or something - it wasn't in the normal flow of applying a configuration but it does happen, so we need to address it.

Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1
ykuijs: @BrianFarnhill Haven't been able to solve this. Left it as it is, so you can check it out. If you have some solution on how to solve this, let me know.

I have the feeling that this is caused by the fact that the module is imported in the *-TargetResource methode, overwriting the Mocked version of the method.

ykuijs:
@BrianFarnhill Ready for another review.

BrianFarnhill:
Reviewed 2 of 7 files at r1, 4 of 4 files at r4, 1 of 1 files at r5.
Review status: all files reviewed at latest revision, 4 unresolved discussions.

---

*[Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1, line 62 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9f9yi7as-lMuw4yUA:-KE9f9yi7as-lMuw4yUB:-1297623855) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1#L62)):*
I feel like we should be reporting the cache account policy status here, just passing through the parameter could give the impression that they have been set to someone who calls this through Get DSC methods so we need to shift the logic to give a true/false indicator here about whether they are configured or not.

---

*[Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1, line 314 \[r2\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE2vphwnItwOEngvxH9-r2-314:-KE9i09Oy3lxwY0FBJVB:-2061007934) ([raw file](https://github.com/powershell/xsharepoint/blob/70d43c3fc3568deaa32abbbcded64c8ba8e3c7a7/Modules/xSharePoint/Modules/xSharePoint.Util/xSharePoint.Util.psm1#L314)):*
I couldn't get this to work any other way either so happy to leave it for now and we can extend it later

---

*[Modules/xSharePoint/Modules/xSharePoint.WebAppPolicy/xSPWebAppPolicy.psm1, line 1 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9ixsUkGiZ5NRs6COQ:-KE9ixsUkGiZ5NRs6COR:658356660) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Modules/xSharePoint/Modules/xSharePoint.WebAppPolicy/xSPWebAppPolicy.psm1#L1)):*
Can I get you to update the name of this file to match the directory name (so xSharePoint.WebAppPolicy.psm1) so that it is inline with our other child modules?

---

*[Tests/xSharePoint/xSharePoint.xSPWebAppPolicy.Tests.ps1, line 1478 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9iYxoJIgz4bK3YiBh:-KE9iYxpBOzjkik_aadq:-1472672793) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Tests/xSharePoint/xSharePoint.xSPWebAppPolicy.Tests.ps1#L1478)):*
Need to add a blank empty line at the end of this file

---


*Comments from the [review on Reviewable.io](https://reviewable.io:443/reviews/powershell/xsharepoint/233)*

ykuijs:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1, line 70 \[r2\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE2vphwnItwOEngvxH7-r2-70:-KE_oB_4aupJYkMiudaW:1724727577) ([raw file](https://github.com/powershell/xsharepoint/blob/70d43c3fc3568deaa32abbbcded64c8ba8e3c7a7/Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1#L70)):*
I have rewritten the code, it is now doing what you are describing.

---

*[Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1, line 62 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9f9yi7as-lMuw4yUA:-KE_njyqqz4Xso1IkeWy:63124538) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1#L62)):*
If we do that, we are implementing a validation test in the Get-TargetResource method. What I am doing now is returning the values and then having Test-TargetResource do the checking. If the SetCacheAccountsPolicy is set to true, the Test method will check if these are in the Web App Policy and return false it they aren't.

---

*[Modules/xSharePoint/Modules/xSharePoint.WebAppPolicy/xSPWebAppPolicy.psm1, line 1 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9ixsUkGiZ5NRs6COQ:-KE_ooBQkD95T63CGtCO:-1612155540) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Modules/xSharePoint/Modules/xSharePoint.WebAppPolicy/xSPWebAppPolicy.psm1#L1)):*
Or course no problem. However, I have used the same format as the other helper modules:
Folder name: <module name>.<resource name>, file name: <resource friendly name>
Folder name: xSharePoint.WebApplication, file name: xSPWebApplication.BlockedFileTypes
Folder name: xSharePoint.WebAppPolicy, file name: xSPWebAppPolicy

Let me know what you prefer!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xsharepoint/233)*

BrianFarnhill:
Review status: all files reviewed at latest revision, 3 unresolved discussions.

---

*[Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1, line 62 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9f9yi7as-lMuw4yUA:-KEcOAFMITw9eYmGTPsp:1222622486) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1#L62)):*
Not really - if we are doing that we are identifying whether or not the cache accounts are set (returning the current status, the purpose of the get method), not whether or not they are in the desired state (checking that the value there matches the desired state, the purpose of the test function). Simply passing the value through in the get method gives me no actual information about whether or not they have been set if I'm looking purely at the output of the get method, which is why this needs to change.

---

*[Modules/xSharePoint/Modules/xSharePoint.WebAppPolicy/xSPWebAppPolicy.psm1, line 1 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9ixsUkGiZ5NRs6COQ:-KEcOozGDXNocC-6mhsr:-840150691) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Modules/xSharePoint/Modules/xSharePoint.WebAppPolicy/xSPWebAppPolicy.psm1#L1)):*
Sorry I glanced wrong at this - happy to go as you have it.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xsharepoint/233)*

ykuijs:
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1, line 62 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9f9yi7as-lMuw4yUA:-KEeeYOcksOmGcmV2NXa:663138082) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1#L62)):*
Ok, sounds logical. Will update my code and push it to the PR

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xsharepoint/233)*

ykuijs:
@BrianFarnhill, ready for final review

ykuijs:
Review status: 3 of 6 files reviewed at latest revision, 2 unresolved discussions.

---

*[Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1, line 62 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9f9yi7as-lMuw4yUA:-KEfj22QMYCUgqK9G9vY:-499146009) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Modules/xSharePoint/DSCResources/MSFT_xSPWebAppPolicy/MSFT_xSPWebAppPolicy.psm1#L62)):*
Done.

---

*[Tests/xSharePoint/xSharePoint.xSPWebAppPolicy.Tests.ps1, line 1478 \[r5\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KE9iYxoJIgz4bK3YiBh:-KEfilMYqa8yqrgIyYoF:-499146009) ([raw file](https://github.com/powershell/xsharepoint/blob/560eacd8a3ce2bf4948193b91db9292fafa63dcb/Tests/xSharePoint/xSharePoint.xSPWebAppPolicy.Tests.ps1#L1478)):*
Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xsharepoint/233)*

BrianFarnhill:
Reviewed 3 of 3 files at r6.
Review status: all files reviewed at latest revision, 2 unresolved discussions.

---

*[README.md, line 131 \[r6\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KEhiMoN2X1W2_CVAvGa:-KEhiMoN2X1W2_CVAvGb:2011441385) ([raw file](https://github.com/powershell/xsharepoint/blob/bc72fb1e72e474b4e44341f004a2b1d812c48e17/README.md#L131)):*
I think this was meant to go under the "unreleased" bit, not under 0.12

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xsharepoint/233)*

ykuijs:
Review status: all files reviewed at latest revision, 1 unresolved discussion.

---

*[README.md, line 131 \[r6\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KEhiMoN2X1W2_CVAvGa:-KEjztehLebAm6dX6mgt:-24995043) ([raw file](https://github.com/powershell/xsharepoint/blob/bc72fb1e72e474b4e44341f004a2b1d812c48e17/README.md#L131)):*
I didn't add this. Aren't you confusing two different reviews??

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xsharepoint/233)*

BrianFarnhill:
Reviewed 1 of 1 files at r7.

---

*[README.md, line 131 \[r6\]](https://reviewable.io:443/reviews/powershell/xsharepoint/233#-KEhiMoN2X1W2_CVAvGa:-KEmn7j3lqsq0TMT9bbP:778780096) ([raw file](https://github.com/powershell/xsharepoint/blob/bc72fb1e72e474b4e44341f004a2b1d812c48e17/README.md#L131)):*
My bad - looked at it wrong in reviewable. Note to self - no more code reviewing before coffee!

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/xsharepoint/233)*

PR #234, opened 2016-03-23
Fixed bug with claims accounts in xSPCacheAccounts

BrianFarnhill:
@ykuijs - Seeing as we discussed this one can I get you to review this one?

### DIFF
--- a/Modules/xSharePoint/DSCResources/MSFT_xSPCacheAccounts/MSFT_xSPCacheAccounts.psm1
+++ b/Modules/xSharePoint/DSCResources/MSFT_xSPCacheAccounts/MSFT_xSPCacheAccounts.psm1
@@ -4,9 +4,10 @@ function Get-TargetResource
     [OutputType([System.Collections.Hashtable])]
     param
     (
-        [parameter(Mandatory = $true)] [System.String] $WebAppUrl,
-        [parameter(Mandatory = $true)] [System.String] $SuperUserAlias,
-        [parameter(Mandatory = $true)] [System.String] $SuperReaderAlias,
+        [parameter(Mandatory = $true)]  [System.String]  $WebAppUrl,
+        [parameter(Mandatory = $true)]  [System.String]  $SuperUserAlias,
+        [parameter(Mandatory = $true)]  [System.String]  $SuperReaderAlias,
+        [parameter(Mandatory = $false)] [System.Boolean] $SetWebAppPolicy = $true,
         [parameter(Mandatory = $false)] [System.Management.Automation.PSCredential] $InstallAccount
     )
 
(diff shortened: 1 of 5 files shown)


Modules/xSharePoint/DSCResources/MSFT_xSPCacheAccounts/MSFT_xSPCacheAccounts.psm1
```diff
@@ -4,9 +4,10 @@ function Get-TargetResource
     [OutputType([System.Collections.Hashtable])]
     param
     (
-        [parameter(Mandatory = $true)] [System.String] $WebAppUrl,
-        [parameter(Mandatory = $true)] [System.String] $SuperUserAlias,
-        [parameter(Mandatory = $true)] [System.String] $SuperReaderAlias,
+        [parameter(Mandatory = $true)]  [System.String]  $WebAppUrl,
+        [parameter(Mandatory = $true)]  [System.String]  $SuperUserAlias,
+        [parameter(Mandatory = $true)]  [System.String]  $SuperReaderAlias,
+        [parameter(Mandatory = $false)] [System.Boolean] $SetWebAppPolicy = $true,
```
ykuijs: Do you really want to default this parameter to True? I would suggest to mark it as False, so it is a conscious decision to change the web app policy.

Modules/xSharePoint/DSCResources/MSFT_xSPCacheAccounts/MSFT_xSPCacheAccounts.psm1
```diff
@@ -90,16 +127,17 @@ function Test-TargetResource
     [OutputType([System.Boolean])]
     param
     (
-        [parameter(Mandatory = $true)] [System.String] $WebAppUrl,
-        [parameter(Mandatory = $true)] [System.String] $SuperUserAlias,
-        [parameter(Mandatory = $true)] [System.String] $SuperReaderAlias,
+        [parameter(Mandatory = $true)]  [System.String]  $WebAppUrl,
+        [parameter(Mandatory = $true)]  [System.String]  $SuperUserAlias,
+        [parameter(Mandatory = $true)]  [System.String]  $SuperReaderAlias,
+        [parameter(Mandatory = $false)] [System.Boolean] $SetWebAppPolicy = $true,
```
ykuijs: Do you really want to default this parameter to True? I would suggest to mark it as False, so it is a conscious decision to change the web app policy.

ykuijs:
How should the cache account users be added to the Web Application Policy?? In domain\users format or Claims format or maybe both?

Reason I ask: In the WebAppPolicy resource I add the user as configured in the web app properties, so not in claims format.

Modules/xSharePoint/DSCResources/MSFT_xSPCacheAccounts/MSFT_xSPCacheAccounts.psm1
BrianFarnhill: I do because that's the behavior we have currently and I don't want to change the behavior for people who are already using xSharePoint. The ones who want to manage the web app policy through the new resource can implement this property to set it to false and get what they want as a behavior moving forward.

BrianFarnhill:
@ykuijs We add them ad DOMAIN\user by default, and then if the web app uses claims we add them as the claims format as well. We have found there are issues when you try to create an enterprise search center in a web app that doesn't have the claims policies enabled, so we end up with both. But the logic is there to do this.

ykuijs:
According to the following article, the account should be entered in the WebApp properties in claims or classic format and then equally in the WebApp policy:
https://technet.microsoft.com/en-us/library/ff758656.aspx
First Step 14 in To create the user accounts by using Central Administration
Then Step 3 in To add the user accounts to the web application by using Windows PowerShell

So when the web app is in Claims mode, the account in the Cache Accounts should be entered in Claims format. So right now this resource is incorrect.

BrianFarnhill:
@ykuijs As per other comment in your PR I'll sort out the fix on this one.

Issue #248, opened 2016-03-29
xSPUsageApplication Get-TargetResource hash table not aligned with mof definition


BrunoReboul:
Error 4252 on Get-TargetResource due to 2 field in the returned hash table 'DatabasePassword' and 'DatabaseUsername' that are missing in the mof definition of xSPUsageApplication.

Proposed solution : remove the 2 fields from the hashtable, (IMO sending back the password iin a string not desirable). Solution tested / works well

Alternate solution, update the mof definition to be aligned.

BrianFarnhill:
Thanks for reporting this - there was a change a while back that swapped those two fields for the PSCredential object $DatabaseCredentials that is there now, and it looks like we missed that hashtable. We'll drop those from the results and replace with the DatabaseCredentials property that will just pass through the PSCredential it's passed in (if any)

BrianFarnhill:
The fix for this is in #253 - I'll close this when the change hits dev

BrianFarnhill:
Just merged this in, closing this bug

PR #253, opened 2016-04-01
Fixed output of get method for xSPUsageApplication

BrianFarnhill:
@caadam - another quickie mate, fix for #248

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/xsharepoint/253)


camiloborges:
all good with this one. Only Remark. This means we are breaking backwards compatibility. 
Should we highlight there is a breaking change?

BrianFarnhill:
@camiloborges no breaking change here, we are actually fixing the output to match the schema - the only time we make a breaking change if if the schema.mof includes a change to a required field (either adding one or modifying one). Everything in the script file is very unlikely to break an existing configuration. I'll merge this one in.

camiloborges:
Hey.
Sorry, My Bad.
I think I was tricked by GIT. I hadn't opened the entire file and the diff
window made me think you had removed the parameters DatabaseName and
DatabaseServer.

PR #320, opened 2016-06-09
Case sensitivity issues after rename

BrianFarnhill:
We had some issues with case sensitivity in the rename (SharePointDSC vs SharePointDsc), this PR resolves those issues


BrianFarnhill:
A current issue has appveyor not connected here, so I ran the tests in my repo to ensure it passed before merging this in. Results at https://ci.appveyor.com/project/BrianFarnhill/sharepointdsc/build/1.0.0.4

PR #351, opened 2016-06-29
Fix issue with group names in claims tokens

### LINKED ISSUE #348: AD Groups don't pass test methods for SPWebAppPolicy and SPServiceAppSecurity

Currently after a change to 1.1 to add support for groups to both SPWebAppPolicy and SPServiceAppSecurity instead of the group name being returned from the "get" methods, the SID for the group is instead returned, causing the test methods to fail and attempt to add the group again. This doesn't appear to cause any issues in terms of permissions, but presents an incorrect "false" result when the system is checked with Test-DscConfiguration. 

**The DSC configuration that is using the resource:**

Any version of SPWebAppPolicy and SPServiceAppSecurity that uses and AD group in their member based properties

**Version of the Operating System and PowerShell the DSC Target Node is running:**

n/a

1.1

**Other notes**

The below example shows how a SID can be converted to the account name of the group, so it just needs to be prefixed with the domain prefix. 

    ([wmi]"Win32_SID.SID='s-1-5-21-2753725054-2932589700-2007370523-2138'").AccountName

BrianFarnhill:
Fixes #348 

- [X] Change details added to Unreleased section of readme.md?
- [n/a] Added/updated documentation and descriptions in .schema.mof files where appropriate?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/351)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Change log for SharePointDsc
 
+### Unreleased
+
+ * Fixed bugs SPWebAppPolicy and SPServiceApPSecurity that prevented the get methods from returning AD group names presented as claims tokens
+
 ### 1.1
 
  * Added SPBlobCacheSettings, SPOfficeOnlineServerBinding, SPWebAppPermissions, SPServiceAppProxyGroup, SPWebAppProxyGroup and SPUserProfileServiceAppPermissions resources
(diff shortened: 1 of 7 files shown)


ykuijs:
Reviewed 7 of 7 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/351)*

PR #371, opened 2016-07-13
Added HostHeader parameters to Web application examples

ykuijs:
Added HostHeader parameters to Web application examples. This way multiple web applications can be created, without creating a blocker in IIS. Without this parameter, the website in IIS is created without host header and all other web applications won't be created due to an error.

- [N/A] Added/updated documentation and descriptions in .schema.mof files where appropriate?
- [N/A] New/changed code adheres to [Style Guidelines]?(https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md)?
- [N/A] [Unit and Integration tests](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md) created/updated where possible?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/371)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,7 @@
  * Fixed the use of plural nouns in cmdlet names within the module
  * Removed dependency on Win32_Product from SPInstall
  * Added SPTrustedIdentityTokenIssuer and SPRemoteFarmTrust resources
+ * Added HostHeader parameter in examples for Web Application, so subsequent web applications won't error out 
 
 ### 1.1
 
--- a/Modules/SharePointDsc/Examples/Single Server/SharePoint.ps1
+++ b/Modules/SharePointDsc/Examples/Single Server/SharePoint.ps1
@@ -173,6 +173,7 @@ Configuration SharePointServer
                 DatabaseName           = $webApp.DatabaseName
                 DatabaseServer         = $ConfigurationData.NonNodeData.SQLServer.ContentDatabaseServer
                 Url                    = $webApp.Url
+                HostHeader             = $webApp.HostHeader
                 Port                   = [Uri]::new($webApp.Url).Port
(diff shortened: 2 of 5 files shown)


BrianFarnhill:
Can I get some info on what the error is that you saw here? I feel like if HostHeader is going to stay optional we should make sure the error doesn't happen rather than just add it to the examples. 

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/371#-:-KMaV3PPB0bqiad-qdUR:-92715693)*

ykuijs:
In the current example, the created web application is created within IIS with a binding of ":80". The current example now only creates one web application, which works fine. But if you add another web application and don't use the HostHeader parameter, the second web application is also created with binding ":80" which is not possible so an error is thrown.

To illustrate the use of this parameter, I have added the parameter to the example.

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/371#-:-KMc6JN7AaPJw-a5eFOx:1664231556)*

BrianFarnhill:
Makes sense - I'll process this one in the morning but I think we need to handle this differently somehow to help steer customers away from doing this. Give it some thought and come back to me and we can have a chat about it

---

Review status: 0 of 5 files reviewed at latest revision, all discussions resolved.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/371#-:-KMcpg7-N9-OiCAMkGL4:-719198788)*

BrianFarnhill:
Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/371)*

PR #376, opened 2016-07-26
Prevented SPCreateFarm and SPJoinFarm from executing sets when in a farm

### LINKED ISSUE #375: SPCreateFarm and SPJoinFarm do not gracefully handle new farm config DB values

Currently both SPCreateFarm and SPJoinFarm will not allow a server to be moved from one farm to another. This means that if a configuration is pushed to a node that has a FarmConfigDatabaseName value that is different to the current value, it will fail the test and attempt to run the set method which blindly does the work without checking status first. Both of these resources should check the current status first and throw a meaningful error in the event that a server is already a member of a different farm.

**Version of the Operating System and PowerShell the DSC Target Node is running:**

Windows Server 2012 R2, WMF 5

SharePointDsc 1.1

BrianFarnhill:
Fixes #375 - we do a check in the set methods now to ensure we aren't trying to create or join where an existing farm is there, warning the user that they must manually remove it first if the config is to be applied.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/376)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -13,7 +13,8 @@
  * Fixed the use of plural nouns in cmdlet names within the module
  * Removed dependency on Win32_Product from SPInstall
  * Added SPTrustedIdentityTokenIssuer, SPRemoteFarmTrust and SPSearchResultSource resources
- * Added HostHeader parameter in examples for Web Application, so subsequent web applications won't error out 
+ * Added HostHeader parameter in examples for Web Application, so subsequent web applications won't error out
+ * Prevented SPCreateFarm and SPJoinFarm from executing set methods where the local server is already a member of a farm 
 
 ### 1.1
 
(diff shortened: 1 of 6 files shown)


BrianFarnhill:
Pushing this bug fix through immediately based on testing I've done with a customer to ensure it is included in the v1.2 release

---

Reviewed 5 of 5 files at r1.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/376#-:-KNdL0s9Ad-ZzVJIi8Ep:-1905036791)*

BrianFarnhill:
Reviewed 1 of 1 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/376)*

PR #350, opened 2016-06-28
Type name in service application resources replaced by GetType Method

luigilink:
Thanks for submitting a Pull Request to this project.

Please make sure you have read the [Contribution Guidelines](https://github.com/powershell/xSharePoint/wiki/Contributing%20to%20xSharePoint).

To aid community reviewers in reviewing and merging your PR, please take the time to run through this checklist:
- [ ] Change details added to Unreleased section of readme.md?
- [x] Added/updated documentation and descriptions in .schema.mof files where appropriate?
- [x] Examples updated for both the single server and small farm templates in the examples folder?
- [x] New/changed code adheres to [Style Guidelines]?(https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md)?
- [x] [Unit and Integration tests](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md) created/updated where possible?

Your contribution to this project is greatly appreciated.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/350)

PR #385, opened 2016-08-08
[Updated] Type name in service application resources replaced by GetType Method

### LINKED ISSUE #342: TypeName in Service Application resources should not use localised strings
**Context:** Windows Server + Language Pack

**Symptom:** Set works, then Test fails, triggering a new Set that end up in error

**Why:** Get function have English Sharepoint name hardcoded. As Sharepoint on a localized OS send back service names in the local language the DSC test functions wrongly determine Ensure as "Absent" while it should be "Present"

**Proposed fix** : update resources to rely on the Object model instead of the TypeName to identify SharePoint Services / objects.

BrianFarnhill:
Replaces #350 
Fixes #342 

Added a lookup that will transition the display names of service instances to the type names for the checks. This does still assume that the language being used is English (a wider strategy needs to be put in place for supporting multiple langauges in SharePointDsc, so happy to discuss as appropriate) but should ensure that consistent use of the English strings will get the services running even when the language pack modifies it under the hood.  

The same unit tests are used to ensure we didn't break backwards compatibility with this change. 

- [n/a] Added/updated documentation and descriptions in .schema.mof files where appropriate?


ykuijs:
Reviewable button is missing, so reviewing the old fashioned way.

Changes look good.

PR #387, opened 2016-08-12
Fix sp search topo get arr string

BrunoReboul:
**Issue:** Get on SPSearchTopology fails on a single serveur farm with error "A general error occurred that is not covered by a more specific error code"

**Root Cause:** multiple setting in the returned hashtable are string instead of expected array of string

**Fix:** reworked code with gettype() the check string and transform to array of string

- [YES ] Change details added to Unreleased section of changelog.md?
- [ NA] Added/updated documentation and descriptions in .schema.mof files where appropriate?
- [ NA] Examples updated for both the single server and small farm templates in the examples folder?
- [ YES] New/changed code adheres to [Style Guidelines]?(https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md)?
- [ YES] [Unit and Integration tests](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md) created/updated where possible?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/387)


BrianFarnhill:
Reviewed 1 of 6 files at r3.
Review status: 1 of 6 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[Modules/SharePointDSC/DSCResources/MSFT_xSPInstallCumulativeUpdate/MSFT_xSPInstallCumulativeUpdate.psm1, line 1 \[r3\]](https://reviewable.io:443/reviews/powershell/sharepointdsc/387#-KP7xYohB_KH5aIt6pNg:-KP7xYoiE50mgr0PorJj:bv5gjqp) ([raw file](https://github.com/powershell/sharepointdsc/blob/1182ddce6fd06a16d2db13f377c8b4fd1f15671d/Modules/SharePointDSC/DSCResources/MSFT_xSPInstallCumulativeUpdate/MSFT_xSPInstallCumulativeUpdate.psm1#L1)):*
> ```PowerShell
> function Get-TargetResource
> ```

I need you to take these CU related ones out of this PR and put them in to a separate PR if they are ready

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/387)*

BrianFarnhill:
Reviewed 1 of 6 files at r3.
Review status: 2 of 6 files reviewed at latest revision, 2 unresolved discussions, some commit checks failed.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPSearchTopology/MSFT_SPSearchTopology.psm1, line 50 \[r3\]](https://reviewable.io:443/reviews/powershell/sharepointdsc/387#-KPABICGDuoWuvuY1U2P:-KPABICGDuoWuvuY1U2Q:bq3gao0) ([raw file](https://github.com/powershell/sharepointdsc/blob/1182ddce6fd06a16d2db13f377c8b4fd1f15671d/Modules/SharePointDsc/DSCResources/MSFT_SPSearchTopology/MSFT_SPSearchTopology.psm1#L50)):*
> ```PowerShell
>         $ConfirmPreference = 'None'
> 
>         $Configuration = @{
> ```

The changes you've made here are way more complicated than they need to be to ensure that arrays come back out of the get methods. You can simply define the variable as an array first and then add to it instead of taking the one value out. So an example of how to do this would be to take the AdminComponents bit on line 72 (before the change) and do something like this with it:

        $AdminComponents = @()
        $AdminComponents += ($allComponents | Where-Object -FilterScript { 
                                ($_.GetType().Name -eq "AdminComponent") 
                            }).ServerId | ForEach-Object -Process {
                                $serverId = $_
                                $server = $allServers | Where-Object -FilterScript {
                                    $_.Id -eq $serverId 
                                } | Select-Object -First 1
                                return $server.Name
                            }

This way it's an array to begin with and stays an array when we add the additional values to it. Even if it's just the one value it will still come back as an array. 

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/387)*

Issue #424, opened 2016-10-11
Infer UseSSL value for SPWebApplication from URL


BrianFarnhill:
**Details of the scenario you try and problem that is occurring:**

Currently [SPWebApplication](https://github.com/PowerShell/SharePointDsc/wiki/SPWebApplication) requires you to set the "UseSSL" property manually, and if you set this to true but provide a URL property that does not start with "https" SharePoint will throw an error. We should be able to detect the type of URL that is provided and automatically set the UseSSL attribute based on that, taking the need to do this away from the author and removing the chance of it being mistakenly entered wrong.

**The DSC configuration that is using the resource:**

```
SPWebApplication SSLFailingWebApp
{
    Url = "http://sharepoint.contoso.com"
    UseSSL = $true
}
```

**Version of the Operating System and PowerShell the DSC Target Node is running:**

n/a

**Version of the DSC module you're using:**

SharePointDsc 1.3

ykuijs:
I am on it!

MikeLacher448:
I had a use case today where my customer wanted to offload SSL to an F5. The way the F5 proxying works, I had to name the web application URL with https, even though it use SSL was set to false, and used port 80. Otherwise, the 302 redirects coming from SharePoint used http in the redirect locations.

BrianFarnhill:
@MikeLacher448 I think we need to check this, because I'm actually pretty sure if you try to create a web app with a https address and you don't use the UseSSL flag it gives an error - we really only rely on this when creating the web apps. Happy to explore the combinations of what New-SPWebApplication will take and make sure we don't exclude any scenarios though. Thanks for raising this!

MikeLacher448:
confirmed, you do get an error trying to create or extend a web application with https and not specifying usessl. In my particular case, the web app had to be created with http://, with the public URLs updated afterwards. This particular case will cause the consistency checks to fail as they check for a http:// web application and don't find it.

BrianFarnhill:
@MikeLacher448 Ahhh, ok so at this point we know things go south in terms of DSC when you start modifying the default zone URLs for the web app, so we don't recommended doing that if you want to manage things with DSC.

ykuijs:
Implemented change for a 2.0 PR

ykuijs:
Fix merged into 2.0 branche

Issue #456, opened 2016-11-13
Breaking changes for a v2.0 release


BrianFarnhill:
I wanted to open an issue that we could link all items that relate to breaking changes that will go in a v2.0 release, as well as have a central place to discuss other items relating to a 2.0 release such as:

- How to document and communicate the breaking changes
- How to work on the breaking changes at the same time (such as a v2.0 dev branch)

I want to group as many of the breaking changes together in to a single release as I can so we don't have a continuing run of them to implement over several months, just a single release with all of the changes in one go will be easier for an administrator to cope with I feel (let me know if you disagree though). 

To get things started, the pending list of breaking changes to be included in a v2.0 release would include the following issues:

- Create SPFarm resource to supercede SPCreateFarm and SPJoinFarm ( #455)
- Infer UseSSL value for SPWebApplication from URL (#424)

I also want to make some changes to search topology and distributed cache as well which I don't have open issues on yet, so I'll get my thoughts on those written down and linked in here as well

BrianFarnhill:
The plan will be to move any breaking change in to the "[preview-2.0](https://github.com/PowerShell/SharePointDsc/tree/preview-2.0)" branch and when we are ready to ship version 2.0 with all of the pending breaking changes we can ship that branch in to dev and go from there.

mrpullen:
#538

ykuijs:
More breaking changes which will be implemented in v2.0:
#644 #423 #550

ykuijs:
PR #660 is implementing most of these breaking changes

ykuijs:
Only issue #538 is still open, all other breaking changes are fixed.

ykuijs:
All breaking issues have been implemented and preparations for the v2.0 release are started

NikCharlebois:
![image](https://user-images.githubusercontent.com/2547149/33721238-6ece9750-db34-11e7-9644-e14b10366f63.png)

ykuijs:
v2.0 has been released, closing issue

PR #459, opened 2016-11-16
Infer the use of SSL

ykuijs:
As suggested in issue #424, this PR implements inferring the use of SSL instead of using the UseSSL parameter

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/459)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,8 @@
   to all be the same case
 * Fixed bug in SPInstallLanguagePack where language packs could not be installed
   on SharePoint 2016
+* Changed SPWebApplication to infer the use of SSL based on the URL instead of the
+  UseSSL parameter
 
 ## 1.4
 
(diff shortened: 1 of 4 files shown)


BrianFarnhill:
@ykuijs Since this is a breaking change, I'm going to hold off on it for a moment, pending what we come up with in the discussion about v2.0 stuff in #456 - I suspect we'll need to merge this to a new branch though.

ykuijs:
What do you mean with a breaking change? I thought that you meant you planned to have them fixed before v2

BrianFarnhill:
@ykuijs By removing the UseSSL property we will break peoples schemas, there is no way around that, so this needs to be held off until we are ready to commit to releasing things as v2.0 - and what I raised as a point in that discussion is that it might be worth grouping all the breaking changes we have pending in issues in to one big release rather than releasing them one and a time.

camiloborges:
Hello. Can I suggest we add a tag 'breakingchange' so it is easy to track these down?

BrianFarnhill:
@ykuijs Can you please re-open this one to merge in to the "[preview-2.0](https://github.com/PowerShell/SharePointDsc/tree/preview-2.0)" branch? We'll park any breaking changes there until we drop 2.0. We can still review this now and merge it in though.

Issue #398, opened 2016-09-01
New improvements to make it usable for automatic installation


paololec:
Please add more details documentations/how-to on wiki (example: guide to automatically install SharePoint 2016 on 4 vm in h.a. + 2 vm for sql and 2 vm for owa/oos: prepare sql, install module, ability this, launch this...) and example of configurations and options to use it.

I use autospinstaller , "SharePointDsc" have same or more features and simple configuration as it? 
I want be able to easy define and do the same thing, but it seems very hard to do it now respect to autospinstaller. 
And autospinstaller is easy debuggable and fixable in case of problems, SharePointDsc is more complicated (or not easy to learn).

The logical step must be (with a GUI as autospinstaller.com ? ):
-define number of server
-define account
-define common settings (web application, version of SharePoint, database name, central admin)
-define services application and role for every server (ex. put distributed cache and search in this server, but frontend here, but application here with this services)

BrianFarnhill:
Hi @paololec - I've been in conversation with @brianlala in regards to how we get customers to transition from AutoSPInstaller to SharePointDsc. We have some ideas in this regard, but nothing that we are ready to start implementing as yet. So if you are dependent on a GUI to create your SharePoint farm, then right now AutoSPInstaller is a better option. I'm hopeful this will change in future, but until we have some specific plans on that front I can't give you a better option.

I will however leave this issue open so we can have some discussion around this issue and as we get closer to implementing a GUI to create configurations with we can refer to this issue.

MikeLacher448:
My 2 cents on this... 
I agree that we can always work on improving documentation and expand on the examples. However, the real benefits of SharePointDsc is not in providing a master script or workflow for everyone to use. Everyone has different requirements on how their farm is built. But rather, SharePointDsc provides the building blocks for you to implement within your process, as simple or as complicated as needed. With SharePointDsc, you can focus on building the farm topology and configuration as code, maintaining it under source control, without focusing on the back end powershell needed to create a farm, create a web application, configure search, etc...

monkeymagik:
I would happily continue using AutoSPInstaller to deploy SharePoint environments, except for the way it handles passwords which DSC does a better and more secure job of and a few little things here and there that I've had to write extra functions.

While it's true every SharePoint install is different, AutoSPInstaller will pretty much do everything you need for 90% of SharePoint environments I've seen.

[...]

PR #514, opened 2017-02-22
AutoSPInstaller Example Code

mrpullen:
#398 

Added Example that would take an AutoSPInstaller.xml file and generate a DSC Configuration. This is still in a preliminary stage, but wanted to put in a pull request to solicit feedback on approach before I go too far down the rabbit hole.

- [NA] Added/updated documentation and descriptions in .schema.mof files where appropriate?
- [NA ] Examples updated for both the single server and small farm templates in the examples folder?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/514)


mrpullen:
My bad. I branched from another pull request. I will clean up the MSFT_SPWebApplication updates tomorrow from this branch. Let me know if this fits with the basic idea of getting DSC to work with the SPAutoInstaller, or if you want to refactor to make the entire thing cleaner. It might be nice to "convert" the XML into a $Configuration hash table or something so we would have a standard object to work with, then pass in the object instead of passing in the XML.

BrianFarnhill:
I will review this one when I get a little more time, but gut feeling says we are on the right path. Shooting from the hip a couple of thoughts:

* I think we might need to look at the credential management stuff in a little more detail. I wanna run some tests to make sure they encrypt right
* I think I want to move the folder out of that examples folder, but will give that some thought too
* We will need to look at unit testing for this, but again I want to give that some thought too

I'll get back to a full review ASAP though

---

Reviewed 1 of 4 files at r2.
Review status: 1 of 3 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/514#-:-Kdy-vsZeZOpgMe9Ku_z:b-9tzf9y)*

mrpullen:
Another option, and maybe something that would work better would be to write a function that would take the XML and just generate a PowerShell data file. Then we could just build a standard Configuration that would take the PowerShell data file configuration. The benefit is that once you ran the script you would be back in "native" DSC land. I might work up a quick sample of this as well. We could push all the specifics into a data file like.

Uploaded a quick sample of what I thought that might look like

BrianFarnhill:
@mrpullen I'm not sure about the second one, I think as a middle ground the initial idea of just parsing the xml like you have to start with is the better approach to take on this one.

mrpullen:
Ok. Will finish up on an xml parser.

mrpullen:
I am going to build this out as a set of composite resources. That way each "top level" XML node can be broken out, making testing each specific piece a bit simpler. I am going to close this PR and submit a new one once that's ready. Finishing a project now, but this will be first up on my list once I am clear.

PR #520, opened 2017-03-02
Update SPTrustedIdentityTokenIssuer to give possibility to load certificate from file system, as an alternative to certificate store

Yvand:
Current version of SPTrustedIdentityTokenIssuer can only get signing certificate from the certificate store of the local machine, but in many cases you want to get the signing certificate from its .cer file.
I updated SPTrustedIdentityTokenIssuer to allow to specify either the thumbprint or the file path.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/520)


BrianFarnhill:
Review status: 0 of 5 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 33 at r2](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KeWNgu8uYCMMmP4S186:-KeWNgu9K5s-vzot3IfD:b19nnhs) ([raw file](https://github.com/powershell/sharepointdsc/blob/115c5d8f21f9bf76ea95ef13bfcb0d70f44d2251/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L33)):*
> ```PowerShell
>         [parameter(Mandatory = $true)]
>         [String]
>         $SigningCertificateThumbprintOrFilePath,
> ```

Rather than change this property to cater for both file path and thumbprint, I think we need to separate them out to be two separate properties. This way when we are looking to return the current status we can be clear that what is being returned is the thumbprint, and then in the set method if the thumbprint is empty we can check the file path as a backup. Can I get you to update your approach to work like this and then we can review this one in full. Let me know if you need more info or need a hand with it at all.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520)*

Yvand:
@BrianFarnhill  this also was my idea initially, but then I realized it could be confusing:
- it's not possible to have either 1 or the other mandatory, it's both or none, so I would have to set both as optional.
- people might think they need to set both properties.

At the end I thought it makes more sense to keep 1 mandatory property that has 1 purpose (get signing certificate), but offers 2 different ways to get it.
What do you think?

BrianFarnhill:
@Yvand We have precedends in other resources where we have similar needs to provide either property A or B to make something go (look at things like SPFarmAdmin where its either Members or MembersToAdd), so adding two properties that are both optional and then throw an error if both or none are used is how we have done this previously. Also in terms of what happens if someone provides both we have a precedent for that too in that if you use both InstallAccount and PsDscRunAsCredential on any resource we'll throw an error at you telling you that you need to use just one.

So I think between meaningful errors and the documentation we generate, using 2 properties is the approach we should take here. @ykuijs you got any strong opinions on this either way mate?

ykuijs:
Agree: Making a parameter mandatory is just using Powershell to enforce the property being specified. When using two optional parameters, you have add that logic yourself to make sure they are used in the correct way.

Another example of a resource that does this is SPSearchFileType:
```
if ($Ensure -eq "Present" -and `
    (-not($PSBoundParameters.ContainsKey("MimeType")) -or `
     -not($PSBoundParameters.ContainsKey("Description"))))
{}
```

Yvand:
Sorry for the late reply, I will make the necessary changes to split the parameter in 2 and update the PR

BrianFarnhill:
Reviewed 8 of 8 files at r3.
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 104 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcdrZJF9sTqzu85fgp:-KfcdrZLa0P5cr2D_BWF:b-2nijyi) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L104)):*
> ```PowerShell
>             IdentifierClaim              = $identifierClaim
>             ClaimsMappings               = $claimsMappings
>             SigningCertificateThumbprintOrFilePath = $SigningCertificateThumbprint
> ```

I still see the "OrFilePath" option being shown here, this should include both the new properties (where path will be blank and the thumbprint property will match)

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 176 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcfMnpA3WWUB3lsNlH:-KfcfMnpA3WWUB3lsNlI:b-galn15) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L176)):*
> ```PowerShell
>         if ($CurrentValues.Ensure -eq "Absent")
>         {
>             if ($PSBoundParameters.ContainsKey("SigningCertificateThumbprint") -and `
> ```

Can you repeat this error check in the test method as well please? 

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 199 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcfAKAVkHixiBOatzu:-KfcfAKAVkHixiBOatzv:b-a0ndn3) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L199)):*
> ```PowerShell
>                 if ($params.SigningCertificateThumbprint)
>                 {
>                     Write-Verbose -Message "Getting signing certificate with thumbprint $($params.SigningCertificateThumbprint) from the certificate store 'LocalMachine\My'"
> ```

Can you please wrap this line to make sure that it doesn't exceed 100 characters in length as per style guidelines

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 201 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcek-jJ_fmxE9FaEGS:-Kfcek-kwy9a0kCzMNsC:bo44ymk) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L201)):*
> ^[A-Fa-f0-9]+$

Can you change this regex to be "^[A-Fa-f0-9]{40}$" - then it will validate the length of the cert as well as it being valid characters. 

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 213 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcf0CY4i92O1_WUUrJ:-Kfcf0CY4i92O1_WUUrK:bb05kvd) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L213)):*
> ```PowerShell
>                     if (!$cert) 
>                     {
>                         throw ("Signing certificate with thumbprint $($params.SigningCertificateThumbprint) was not found in certificate store 'LocalMachine\My'.")
> ```

Can you please undo the change to put this on one line? It was wrapped to make the line fall under the style guideline rule of no lines longer than 100 characters

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 219 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcf8eV-z8I4D5ecjyH:-Kfcf8eV-z8I4D5ecjyI:bb05kvd) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L219)):*
> ```PowerShell
>                     if ($cert.HasPrivateKey) 
>                     {
>                         throw ("SharePoint requires that the private key of the signing certificate is not installed in the certificate store.")
> ```

Can you please undo the change to put this on one line? It was wrapped to make the line fall under the style guideline rule of no lines longer than 100 characters

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 228 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcfGxiWrfVhODOMF-q:-KfcfGxiWrfVhODOMF-r:b-a0ndn3) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L228)):*
> ```PowerShell
>                     try
>                     {
>                         $cert = New-Object -TypeName "System.Security.Cryptography.X509Certificates.X509Certificate2" -ArgumentList @($params.SigningCertificateFilePath)
> ```

Can you please wrap this line to make sure that it doesn't exceed 100 characters in length as per style guidelines

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.schema.mof, line 21 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcfd1STHoLH2EpltT7:-Kfcfd1STHoLH2EpltT8:bapydyc) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.schema.mof#L21)):*
> ```
>     [Required, Description("Array of MSFT_SPClaimTypeMapping to use with cmdlet New-SPClaimTypeMapping"), EmbeddedInstance("MSFT_SPClaimTypeMapping")] String ClaimsMappings[];
>     [Write, Description("Specify the thumbprint of the signing certificate, which must be located in certificate store LocalMachine\\My")] String SigningCertificateThumbprint;
>     [Write, Description("Specify the file path to the signing certificate")] String SigningCertificateFilePath;
> ```

"Specify the file path to the signing certificate if it is not stored in the local certificate store already" would be a better description here

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/readme.md, line 8 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcg5fkmqi6iT9as_BO:-Kfcg5fmdtjQkZAWou0W:b-u9lri3) ([raw file](https://github.com/powershell/sharepointdsc/blob/a01c5ad95541ead0f1d607bb7ef2acaeca65f939/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/readme.md#L8)):*
> ```Markdown
> The SigningCertificateThumbPrint must match the thumbprint of a certificate in
> the store LocalMachine\My of the server that will run this resource.
> Note that the private key of the certificate must not be available in the
> ```

Can you make sure you don't remove the note about the private key being necessary here

---

*[Modules/SharePointDsc/Examples/Resources/SPTrustedIdentityTokenIssuer/1-SigningCertInFilePath.ps1, line 34 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcgHrDGj7n34mPAbqO:-KfcgHrDGj7n34mPAbqP:b-8e32jf) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/Examples/Resources/SPTrustedIdentityTokenIssuer/1-SigningCertInFilePath.ps1#L34)):*
> ```PowerShell
>                     }
>                 )
>                 SigningCertificateFilePath   = "F:\Data\DSC\FakeSigning.cer"
> ```

Can you undo this change and leave the thumbprint example as the first, and the file path example as the second one, so the example number stays consistent between versions where we can help it.

---

*[Modules/SharePointDsc/Examples/Resources/SPTrustedIdentityTokenIssuer/2-SigningCertInFileCertificateStore.ps1, line 3 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcgsZ5cOVaHwJMMGeE:-KfcgsZ6iAnSNsFIe9Tl:bp4v3ji) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/Examples/Resources/SPTrustedIdentityTokenIssuer/2-SigningCertInFileCertificateStore.ps1#L3)):*
> ```PowerShell
> <#
> .EXAMPLE
>     This example deploys a trusted token issuer to the local farm.
> ```

As per above comment, lets make this file the path example

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 25 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcgxYY90kZWClIG-po:-KfcgxYY90kZWClIG-pp:b2fhlfb) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L25)):*
> ```PowerShell
>             return @(
>                 @{
>                     Thumbprint = "123ABCFACE"
> ```

After you update the regex above this will need to be a valid thumbprint (so 40 characters long)

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 28 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfchEJgu2_IBaTEEO32:-KfchEJgu2_IBaTEEO33:bcuniuf) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L28)):*
> ```PowerShell
>                 }
>             )
>         } { $Path -eq 'Cert:\LocalMachine\My' } -Verifiable
> ```

We shouldn't need the -Verifiable switch here, and can you also put -ParameterFilter in front of that second script block there as it's intent is confusing without knowing the positional parameters

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 54 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfchTF2FHTJ9dBcdIN2:-KfchTF3fupA9qxqd8ZX:bv8o8e6) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L54)):*
> ```PowerShell
>         $CsharpCode = @"
> namespace Microsoft.SharePoint.Administration {
>     //public enum SPUrlZone { Default };
> ```

Is this commented out line meant to be here? I feel like it's not so can we please remove it

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 63 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfci03j4KHOsCJJaGbd:-Kfci03j4KHOsCJJaGbe:b-vwufqi) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L63)):*
> ```PowerShell
> 
>         # Test contexts
>         Context -Name "SPTrustedLoginProvider is created using a signing certificate in the certificate store" -Fixture {
> ```

Why have you changed the name of this context? I would have expected that for these changes the only changes to be made to any unit tests would have been to add new ones to test the behaviour of your new parameter combinations that are available, changing the previous tests shouldn't be needed (outside of fixing the thumbprint parameter as mentioned elsewhere). I haven't reviewed the rest of the test file here yet based on waiting for a response to this to see if we should proceed in its new form or make changes to go back to what we had with the new tests added.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520)*

Yvand:
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 104 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcdrZJF9sTqzu85fgp:-KffFdRhz-2eWr0-jWKL:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L104)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

I still see the "OrFilePath" option being shown here, this should include both the new properties (where path will be blank and the thumbprint property will match)
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 199 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcfAKAVkHixiBOatzu:-KffG-3AFa_fvIETyZI1:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L199)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Can you please wrap this line to make sure that it doesn't exceed 100 characters in length as per style guidelines
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 201 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcek-jJ_fmxE9FaEGS:-KffGRX-t7an6Bq_K2oc:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L201)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

> ^[A-Fa-f0-9]+$

Can you change this regex to be "^[A-Fa-f0-9]{40}$" - then it will validate the length of the cert as well as it being valid characters. 
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 213 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcf0CY4i92O1_WUUrJ:-KffGqxj4hagijA9oAML:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L213)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Can you please undo the change to put this on one line? It was wrapped to make the line fall under the style guideline rule of no lines longer than 100 characters
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 219 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcf8eV-z8I4D5ecjyH:-KffGrM8AWXwXAY4VoT2:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L219)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Can you please undo the change to put this on one line? It was wrapped to make the line fall under the style guideline rule of no lines longer than 100 characters
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 228 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcfGxiWrfVhODOMF-q:-KffHeJZTxW1WjLOCxIi:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L228)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Can you please wrap this line to make sure that it doesn't exceed 100 characters in length as per style guidelines
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.schema.mof, line 21 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcfd1STHoLH2EpltT7:-KffHvSbP13gu5ZnQrnC:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.schema.mof#L21)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

"Specify the file path to the signing certificate if it is not stored in the local certificate store already" would be a better description here
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/readme.md, line 8 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfcg5fkmqi6iT9as_BO:-KffIrjwvpIOcsBHQTDf:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/a01c5ad95541ead0f1d607bb7ef2acaeca65f939/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/readme.md#L8)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Can you make sure you don't remove the note about the private key being necessary here
</blockquote></details>

Done.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520)*

Yvand:
Review status: all files reviewed at latest revision, 15 unresolved discussions.

---

*[Modules/SharePointDsc/Examples/Resources/SPTrustedIdentityTokenIssuer/1-SigningCertInFilePath.ps1, line 34 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcgHrDGj7n34mPAbqO:-KffihMWhkCSQhaMxton:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/Examples/Resources/SPTrustedIdentityTokenIssuer/1-SigningCertInFilePath.ps1#L34)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Can you undo this change and leave the thumbprint example as the first, and the file path example as the second one, so the example number stays consistent between versions where we can help it.
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/Examples/Resources/SPTrustedIdentityTokenIssuer/2-SigningCertInFileCertificateStore.ps1, line 3 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcgsZ5cOVaHwJMMGeE:-KffiiUBOPFrd8myCtPb:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/Examples/Resources/SPTrustedIdentityTokenIssuer/2-SigningCertInFileCertificateStore.ps1#L3)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

As per above comment, lets make this file the path example
</blockquote></details>

Done.

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 25 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcgxYY90kZWClIG-po:-Kffl9PriC37F8djimTe:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L25)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

After you update the regex above this will need to be a valid thumbprint (so 40 characters long)
</blockquote></details>

Done.

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 54 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfchTF2FHTJ9dBcdIN2:-KfflJUWHGHtbKpWCsVT:b-ksnalt) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L54)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Is this commented out line meant to be here? I feel like it's not so can we please remove it
</blockquote></details>

It's actually intended: if I run Pester for this test script only I need this class, but if I run Pester with all the scripts I must comment out this class, otherwise it Pester will throw an error saying that the type already exists

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 63 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfci03j4KHOsCJJaGbd:-Kffln8BCa21AphhZ9GD:b-k3726l) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L63)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Why have you changed the name of this context? I would have expected that for these changes the only changes to be made to any unit tests would have been to add new ones to test the behaviour of your new parameter combinations that are available, changing the previous tests shouldn't be needed (outside of fixing the thumbprint parameter as mentioned elsewhere). I haven't reviewed the rest of the test file here yet based on waiting for a response to this to see if we should proceed in its new form or make changes to go back to what we had with the new tests added.
</blockquote></details>

When I created the test file Pester was completely new to me, now I get a better sense of how Pester works and I used this change as an opportunity to clean and optimize the test file. Those changes make tests easier to read, for example I now reuse most my mocks instead of recreating them in each context.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520)*

Yvand:
Review status: all files reviewed at latest revision, 16 unresolved discussions.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 33 at r2](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KeWNgu8uYCMMmP4S186:-KffmrVx5B_IRtr4qVW1:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/115c5d8f21f9bf76ea95ef13bfcb0d70f44d2251/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L33)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Rather than change this property to cater for both file path and thumbprint, I think we need to separate them out to be two separate properties. This way when we are looking to return the current status we can be clear that what is being returned is the thumbprint, and then in the set method if the thumbprint is empty we can check the file path as a backup. Can I get you to update your approach to work like this and then we can review this one in full. Let me know if you need more info or need a hand with it at all.
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 176 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcfMnpA3WWUB3lsNlH:-Kffn9PKS4inmsvQA3l7:b-frtwry) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L176)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

Can you repeat this error check in the test method as well please? 
</blockquote></details>

In the Pester test I already added 2 contexts: "Both parameters SigningCertificateThumbprint and SigningCertificateFilePath are set" and "None of parameters SigningCertificateThumbprint and SigningCertificateFilePath is set".
Which test is missing?

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520)*

BrianFarnhill:
Reviewed 1 of 10 files at r3, 8 of 8 files at r4.
Review status: all files reviewed at latest revision, 6 unresolved discussions.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 104 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcdrZJF9sTqzu85fgp:-KfmqT8tShjFzTgpthLp:b-jaaycl) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L104)):*
<details><summary><i>Previously, Yvand (Yvan Duhamel) wrote…</i></summary><blockquote>

Done.
</blockquote></details>

We need to include the path property here as well, even if its just blank so that way the schema matches when it returns. 

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 176 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcfMnpA3WWUB3lsNlH:-Kfmqcsjht3RSYKP6jMu:bp12edd) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L176)):*
<details><summary><i>Previously, Yvand (Yvan Duhamel) wrote…</i></summary><blockquote>

In the Pester test I already added 2 contexts: "Both parameters SigningCertificateThumbprint and SigningCertificateFilePath are set" and "None of parameters SigningCertificateThumbprint and SigningCertificateFilePath is set".
Which test is missing?
</blockquote></details>

No not in the pester tests, in the Test-TargetResource function.

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 54 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfchTF2FHTJ9dBcdIN2:-KfmrG41c0f9mUXHIlXB:b-usz0ml) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L54)):*
<details><summary><i>Previously, Yvand (Yvan Duhamel) wrote…</i></summary><blockquote>

It's actually intended: if I run Pester for this test script only I need this class, but if I run Pester with all the scripts I must comment out this class, otherwise it Pester will throw an error saying that the type already exists
</blockquote></details>

ok thats an issue then and we can't leave it like this - what you will need to do to determine if the type is already loaded is to do a test and load it when its not there. Have a look at [SPWebApplicationExtention's tests](https://github.com/PowerShell/SharePointDsc/blob/dev/Tests/Unit/SharePointDsc/SharePointDsc.SPWebApplicationExtension.Tests.ps1#L23) for an example of how to put this sort of thing in to a try/catch loop

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 63 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfci03j4KHOsCJJaGbd:-Kfmrnt_rLTA57GWnn-w:b-7nbprf) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L63)):*
<details><summary><i>Previously, Yvand (Yvan Duhamel) wrote…</i></summary><blockquote>

When I created the test file Pester was completely new to me, now I get a better sense of how Pester works and I used this change as an opportunity to clean and optimize the test file. Those changes make tests easier to read, for example I now reuse most my mocks instead of recreating them in each context.
</blockquote></details>

While I don't have an issue with you moving mocks outside of specific contexts if they are the same at each call, I do have an issue with the renaming of the contexts. We should be keeping these in the format of "The server is currently like this, but should be like this" with any additional detail after that, as this best describes the context that DSC will execute within. So in the case of the example with this line it should read something like "The SPTrustedTokenIssuer does not exist, but it should be present, and will be created with a certificate from the local store" or something to that extent. This is easier to read and lines up with the naming of our other context blocks across all tests. 

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 236 at r4](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfmtrsRnLDXvDFBveuq:-KfmtrsSkmFXPRAJoLN7:bkb8u8s) ([raw file](https://github.com/powershell/sharepointdsc/blob/ed6376856551bea86a8c6b3d7fe50e74e07ebe71/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L236)):*
> Pricate key

Typo here, but again the name needs to go back to the format we used before hand

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 273 at r4](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfmueUebkUBW-sRSBud:-KfmueUf_R-q8p8qdbYS:bvnmdgs) ([raw file](https://github.com/powershell/sharepointdsc/blob/a01c5ad95541ead0f1d607bb7ef2acaeca65f939/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L273)):*
> ```PowerShell
>         }
> 
>         Context -Name "The certificate thumbprint does not match a certificate in certificate store LocalMachine\My" -Fixture {
> ```

I can't see this context moved anywhere else, did you remove this or have I missed it? If so it needs to be restored

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520)*

Yvand:
Review status: all files reviewed at latest revision, 7 unresolved discussions, some commit checks failed.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 104 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcdrZJF9sTqzu85fgp:-KfzH6ZrT90wNT_DQIOQ:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L104)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

We need to include the path property here as well, even if its just blank so that way the schema matches when it returns. 
</blockquote></details>

Done.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1, line 176 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfcfMnpA3WWUB3lsNlH:-KfzH_dc_mvFgMazG3W8:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Modules/SharePointDsc/DSCResources/MSFT_SPTrustedIdentityTokenIssuer/MSFT_SPTrustedIdentityTokenIssuer.psm1#L176)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

No not in the pester tests, in the Test-TargetResource function.
</blockquote></details>

Done.

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 28 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfchEJgu2_IBaTEEO32:-KfzHtC3NsetG4lTBE-1:bdssqzs) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L28)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

We shouldn't need the -Verifiable switch here, and can you also put -ParameterFilter in front of that second script block there as it's intent is confusing without knowing the positional parameters
</blockquote></details>

I removed "-Verifiable" switch everywhere, I don't know what it means but Pester still completes successfully so I conclude that it's good!

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 54 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfchTF2FHTJ9dBcdIN2:-KfzONJd9_O5vYuPtkE6:b-xzk0kr) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L54)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

ok thats an issue then and we can't leave it like this - what you will need to do to determine if the type is already loaded is to do a test and load it when its not there. Have a look at [SPWebApplicationExtention's tests](https://github.com/PowerShell/SharePointDsc/blob/dev/Tests/Unit/SharePointDsc/SharePointDsc.SPWebApplicationExtension.Tests.ps1#L23) for an example of how to put this sort of thing in to a try/catch loop
</blockquote></details>

Awesome, indeed it's a lot better like this, changes done for both types.

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 273 at r4](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfmueUebkUBW-sRSBud:-KfzMz0yRXPUYX73my_g:b-d2n14i) ([raw file](https://github.com/powershell/sharepointdsc/blob/a01c5ad95541ead0f1d607bb7ef2acaeca65f939/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L273)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

I can't see this context moved anywhere else, did you remove this or have I missed it? If so it needs to be restored
</blockquote></details>

Context "The IdentifierClaim does not match one of the claim types in ClaimsMappings" is the last one

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520)*

Yvand:
Review status: all files reviewed at latest revision, 6 unresolved discussions, some commit checks failed.

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 63 at r3](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-Kfci03j4KHOsCJJaGbd:-KfzTt3qVRcf-luIf-on:b-h4nkzi) ([raw file](https://github.com/powershell/sharepointdsc/blob/fea608c615cc12d040b1944b60703ff25a190072/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L63)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

While I don't have an issue with you moving mocks outside of specific contexts if they are the same at each call, I do have an issue with the renaming of the contexts. We should be keeping these in the format of "The server is currently like this, but should be like this" with any additional detail after that, as this best describes the context that DSC will execute within. So in the case of the example with this line it should read something like "The SPTrustedTokenIssuer does not exist, but it should be present, and will be created with a certificate from the local store" or something to that extent. This is easier to read and lines up with the naming of our other context blocks across all tests. 
</blockquote></details>

Understood, I hope I renamed all of them like you expect, I was inspired by SPWebApplicationExtension

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 236 at r4](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfmtrsRnLDXvDFBveuq:-KfzU6zpWORV0c9EO5ha:b-896fix) ([raw file](https://github.com/powershell/sharepointdsc/blob/ed6376856551bea86a8c6b3d7fe50e74e07ebe71/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L236)):*
<details><summary><i>Previously, BrianFarnhill (Brian Farnhill) wrote…</i></summary><blockquote>

> Pricate key

Typo here, but again the name needs to go back to the format we used before hand
</blockquote></details>

Done.

---

*[Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1, line 273 at r4](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-KfmueUebkUBW-sRSBud:-KfzU9Frqha7II0N1aLh:b-1zm3uu) ([raw file](https://github.com/powershell/sharepointdsc/blob/a01c5ad95541ead0f1d607bb7ef2acaeca65f939/Tests/Unit/SharePointDsc/SharePointDsc.SPTrustedIdentityTokenIssuer.Tests.ps1#L273)):*
<details><summary><i>Previously, Yvand (Yvan Duhamel) wrote…</i></summary><blockquote>

Context "The IdentifierClaim does not match one of the claim types in ClaimsMappings" is the last one
</blockquote></details>

I just renamed it to "The SPTrustedLoginProvider is desired, but the IdentifierClaim parameter does not match a claim type in ClaimsMappings"

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520)*

BrianFarnhill:
:lgtm:

---

Reviewed 3 of 3 files at r5.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/520#-:-KgBwwFJqFYVmoc2qxaa:bnfp4nl)*

PR #495, opened 2017-01-24
Fix MSFT_SPCreateFarm.psm1 and MSFT_SPJoinFarm.psm1

luigilink:
Fix non desired disconnect SPFarm when get-SPFarm return null.

- [x] Change details added to Unreleased section of changelog.md?
- [x] Added/updated documentation and descriptions in .schema.mof files where appropriate?
- [x] Examples updated for both the single server and small farm templates in the examples folder?
- [x] New/changed code adheres to [Style Guidelines](https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md)?
- [x] [Unit and Integration tests](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md) created/updated where possible?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/495)

PR #521, opened 2017-03-04
New resource SPFarm

### LINKED ISSUE #455: Create SPFarm resource to supercede SPCreateFarm and SPJoinFarm
After running around with the v0.x and v1.x versions of SharePointDsc for a while now I've come across an issue with the approach to the create/join farm resources and it is to do with the disaster recovery scenario for a server that has CreateFarm in its config. Basically it will only ever join a server to a farm when there is no farm there. If you decide you need to destroy and re-create a server that uses SPCreateFarm it will never succeed as it will never be able to create another farm, so you need to manually re-join it to the existing farm for the resource to pass and continue it's configuration.

Based on that, there really should be just a single resource called SPFarm which will create the farm if it doesn't exist, and if it does then join a server to it as well. This will address the above scenario (and be a step towards removing the need for a "farm" server that we have been using at the moment, which creates issues for high availability). I've got some theories about how to implement this which I'll open another issue on as there are wider reaching issues than just these resources, so this issue can focus discussion on the SPFarm resource (which will be a v2.0 resource) and I'll link other related discussions here as I open them.

BrianFarnhill:
Fixes #455 

Also relates to #495 in that I think we can adapt the code there to use the same helper methods I've put in here (would be keen to hear what you think @luigilink - have a look at the SPFarm helper module).

The main difference to the approach though is that I'm creating a "stub" database to use as a lock to make sure multiple servers don't fight over who creates the farm. The stub DB gets ditched at the end of the process, so we aren't relying on creating different DBs at this point as it only exists for a few minutes. Keen to hear what you think @ykuijs 

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/521)


BrianFarnhill:
I need this to be in the 1.6 release, so approving as it's been here for a few days with no comment and we can capture any fixes ahead of a future release if needed.

---

Reviewed 13 of 15 files at r1, 2 of 2 files at r2.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/521#-:-Kef5Yewg62iXH6qIAvr:bnezo1)*

ykuijs:
@BrianFarnhill, you are a little quick, was about to review it. No comments though, it looks great!

Issue #540, opened 2017-03-20
Extend SPOutgoingEmailSettings with TLS and SMTP port number for SP2016


nissan:
Details of the scenario you try and problem that is occurring:
[Help Wanted]
I want to turn TLS Encryption off for my outgoing email settings. [Wiki Documentation for SPOutgoingEmailSettings](https://github.com/PowerShell/SharePointDsc/wiki/SPOutgoingEmailSettings) does not show this as an available parameter, although I see [Powershell commands documented](https://social.technet.microsoft.com/wiki/contents/articles/34167.sharepoint-2016-outgoing-email-configuration-settings.aspx#Configure_using_the_PowerShell) that allows this to be done manually 
Can this be automated presently as a parameter on SPOutgoingEmailSettings? Or does this presently require a manual configuration step?

Attaching the screenshot at the farm level setting that I wish to configure
![web application outgoing email settings](https://cloud.githubusercontent.com/assets/12583/24118191/4b8fcdd8-0d83-11e7-897c-8007efd09842.PNG)
 
The DSC configuration that is using the resource:
Single Server Farm created using SharePointDSC with outgoing email settings configured using the SPOutgoingEmailSettings at the farm level (the central administration web application)

Version of the Operating System and PowerShell the DSC Target Node is running:
Target system is running Window Server 2012 R2 running Powershell v5.0.10586.20151218

Version of the DSC module you're using:
SharePointDSC v1.6.0.0

ykuijs:
Hi @nissan, this functionality has been added in SP2016. The resouces was build when SP2016 wasn't available yet. We will have to extend the functionality of this resource.

ykuijs:
Will be added in my next rollup PR

ykuijs:
Fix included in PR 600

PR #541, opened 2017-03-20
Update to add TLS Encryption Setting as a parameter

nissan:
Add TLS Encryption Setting as a parameter which can be set on ($true) or off ($false) explicitly

- [ ] Change details added to Unreleased section of changelog.md?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/541)

### DIFF
--- a/Modules/SharePointDsc/DSCResources/MSFT_SPOutgoingEmailSettings/MSFT_SPOutgoingEmailSettings.psm1
+++ b/Modules/SharePointDsc/DSCResources/MSFT_SPOutgoingEmailSettings/MSFT_SPOutgoingEmailSettings.psm1
@@ -23,6 +23,10 @@ function Get-TargetResource
         [parameter(Mandatory = $true)]  
         [System.String] 
         $CharacterSet,
+        
+        [parameter(Mandatory =$true)]
+        [System.Boolean]
+        $TLSEncryptionSetting,
 
         [parameter(Mandatory = $false)] 
         [System.Management.Automation.PSCredential] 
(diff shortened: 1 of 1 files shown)


BrianFarnhill:
OK so you're missing quite a bit from this PR in terms of what we need for it to actually work - can I ask if you tested this anywhere first or just submitted the code?

Things that are missing:

* You need to make changes to the Schema.mof file for the new property to be usable at all
* There is no changes to the unit tests to validate the logic on the new property
* You can't make this property mandatory as SharePoint 2013 doesn't have it at all, which means it has to be optional, and in the code we have to do a version check so that if someone supplies this property on SharePoint 2013 we throw an error
* The changelog needs to be updated to add notes on this change
* There needs to be an additional example file added for this resource as well

I'm happy with the basic logic of how you have done the get/set/test of the logic - but we need the rest of the above items to be resolved before we can bring this in to the codebase.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/541#-:-KfnIR1yjmiEoOx9e5Cz:bgmajll)*

ykuijs:
Would be great if you could also add the SMTP port setting. And please also include a check for version of SharePoint, since these settings are only present in SP2016.

---

Review status: 0 of 1 files reviewed at latest revision, all discussions resolved, some commit checks failed.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/541#-:-Kg1bvxHMg2Lsz99-z4M:b-bedtpr)*

ykuijs:
Linking to issue #540

ykuijs:
@nissan, any update on this PR?

BrianFarnhill:
I'm closing this one due to lack of response.

PR #556, opened 2017-03-31
Sp app principal

mrpullen:
Adding a resource for SPAppPrincipal. Want to be able to generate an App Principal, and assign permissions using DSC for some continuous integration work. Next resource would be to deploy an SPAppPackage (side loaded).

- [x] Added/updated documentation and descriptions in .schema.mof files where appropriate?
- [x] Examples updated for both the single server and small farm templates in the examples folder?
- [x] [Unit and Integration tests](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md) created/updated where possible?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/556)


mrpullen:
This still needs some work. Looks like I need to make sure my parameters for SPAppPrincipal permissions are what SharePoint expects, for Scope and Level. The documentation on MSDN isn't great. Will update tomorrow.

ykuijs:
@mrpullen Any update? Your checks are currently failing because of a missing }.

mrpullen:
I will take a look tonight. I got sidetracked and this fell off my radar. Thanks for the ping ;)

BrianFarnhill:
Just a quick one on this (and it's not just you that I'm throwing this one at, just happened to be your PR that I was looking at for it right now), can I ask the question "Have you actually tested this on a SharePoint server?" - because looking at the issues here with the missing } and also the schema failing to compile (which is caused by the use of the ' in the description of the AppId property) there is no way that DSC could have actually run this code for it to work before the PR got submitted. 

I'm always super happy to get contributions from people, and yes the scripts that we get sent usually work fine, but I would like to know that there is at least a chance that stuff is actually being tested before it gets submitted to us for a review.

mrpullen:
So I started this with just the SPAppPrincipal, and no permissions. I will pull this back, sort it and resubmit.

PR #666, opened 2017-10-31
Updated the SPManagedMetaDataServiceApp Resource to allow the edit of the Language Settings

### LINKED ISSUE #664: Managed Metadata Service Application: Modify the default and working languages
In a recent multi language deployment I came across the requirement to configure the managed metadata service application further than the current resources allow (afaik). 

image

*Default language*
In a farm with more than one language installed there is an option to change the default language of the term store.  The default language and the possible options depend on the installed language packs.

*Working languages*
There are two options available for workings languages: "Installed languages" and "Other locales". 

It would be great to have a DSC resource available to set the language preferences.

SharePointDSC 1.9.0.0

andikrueger:
Closes #664 

This resource is responsible for configuring language (default and working) settings for the managed metadata service application. These settings are applied to the managed metadata service for the farm and do not need to be applied to each server individually, the settings will be propagated throughout the farm.

- [x] Added/updated documentation and descriptions in .schema.mof files where appropriate?
- [x] Examples updated for both the single server and small farm templates in the examples folder?
- [x ] [Unit and Integration tests](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md) created/updated where possible?

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/666)


ykuijs:
Hi @andikrueger, thanks for this contribution. I do have one question before I start the review: Why did you choose to create a new resource? Wouldn't it make more sense to add these two additional properties to the "SPManagedMetaDataServiceApp" resource?

Could be I am overlooking a very practical issue here, just wanted to check ;-)

andikrueger:
Hi @ykuijs,
I can't remember exactly why I chose to create a new resource. It felt like the right way to go... Maybe the Search and User Profile resources did influence me in the decision - E.g. There are resources available to configure user profile properties.

Another point, that might have pointed my attention to a new resource:
I did a quick check on the "SPManagedMetadataServiceApp" resource. There is an possible issue, if there is more than one Managed Metadata Service Application available in the farm. The Get-Method assumes the right termstore is always the first one returned in the TermStoreCollection. I think that's why I chose to create the new resource to target the right Term store with the changes that should be applied. That's something that needs to be fixed anyway.

Anyhow from a users point of view: Having just one resource "SPManagedMetaDataServiceApp"  its easier to live
From a dev point of view: The resource is getting more complex and the Get-, Set- and Test- functions get harder to test.

I'm good with either adding these two params to the existing resource and modify the "Get" of the TermStores or keep the new one. I'm fine with any.

ykuijs:
Since there are only two settings, which are also service app wide I agree with you. I like them added to the SPManagedMetadataServiceApp better. It is always good to do a little additional coding to make usability better. Just update the resource and let me know when you are ready for the review.

P.s. Also do an update from the Dev branch. Since I just merged your other PR, this one needs another update to be mergable.

andikrueger:
Currently WIP. 
TODO: add the tests from the new resource to the SPManagedMetadataServiceApp resource tests.

andikrueger:
@ykuijs I would call this the first good looking version. 
I need to do some real world testing, as the changes are fairly large compared to the two new optional parameters. I will look into this further.

andikrueger:
I did some testing against SP2016 and found some issues that need to be addressed:

1. The [DefaultLanguage](https://msdn.microsoft.com/en-us/library/microsoft.sharepoint.taxonomy.termstore.defaultlanguage(v=office.14).aspx) has to be set by a user with TaxonomyRights.ManageTermStore permission. Same applies for the Add and DeleteLanguages methods. The user running the command needs to have ManageTermStore permissions or has to be a member of the Termstore Administrators. 

2. During the first run the Set-TargetResource Method creates the Service Application and the Proxy, but does not apply any changes to Languages, DefaultLanguage or ContentType Hub. These settings are applied in the second run.

3. Some additional tests have to be integrated for 2.

andikrueger:
@ykuijs This check-in looks very stable to me. I'm really looking forward to your review :-)

I did a lot of testing to create, change or delete the service application. It looks very stable now.

I did a bit of refactoring on the Set-TargetResource Method. The remove of parameters to match the New-SPManagedMetadataApplication params was not good. All following ifs failed because the parameters were remove. Therefore I created a new object to pass the required params to the cmdlet.

Additionally I added some handling if PSDscRunAsCredential or install account do not have the required permissions on the termstore to perform the required actions.

Furthermore it is now possible to change the name of the Managed Metadata Service Application proxy.

ykuijs:
Looks good, just one question in the Get method

---

Reviewed 3 of 6 files at r3, 3 of 3 files at r5, 2 of 3 files at r6.
Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPManagedMetadataServiceApp/MSFT_SPManagedMetaDataServiceApp.psm1, line 97 at r6](https://reviewable.io:443/reviews/powershell/sharepointdsc/666#-KyIUNpO5FoFWtDVLyl-:-KyIUNpPA4q3Rebgux0X:b-qfhkq4) ([raw file](https://github.com/powershell/sharepointdsc/blob/81be4f8ab19f4a2885cf192e50c62051585c43e7/Modules/SharePointDsc/DSCResources/MSFT_SPManagedMetadataServiceApp/MSFT_SPManagedMetaDataServiceApp.psm1#L97)):*
> ```PowerShell
>             }
> 
>             if ($null -eq $proxyName)
> ```

Why are you generating the proxy name and not actually retrieving it, like we do it in SPSecureStoreServiceApp

            $serviceAppProxies = Get-SPServiceApplicationProxy -ErrorAction SilentlyContinue
            if ($null -ne $serviceAppProxies)
            {
                $serviceAppProxy = $serviceAppProxies | Where-Object -FilterScript { 
                    $serviceApp.IsConnected($_)
                }
                if ($null -ne $serviceAppProxy) 
                { 
                    $proxyName = $serviceAppProxy.Name
                }
            }

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/666#-:-KyIWBAf9O-p_RUv8m3Q:b-esmvkp)*

ykuijs:
Oh, and please to a pull from Dev to resolve conflicts in the changelog

andikrueger:
Review status: 7 of 8 files reviewed at latest revision, 1 unresolved discussion, some commit checks failed.

---

*[Modules/SharePointDsc/DSCResources/MSFT_SPManagedMetadataServiceApp/MSFT_SPManagedMetaDataServiceApp.psm1, line 97 at r6](https://reviewable.io:443/reviews/powershell/sharepointdsc/666#-KyIUNpO5FoFWtDVLyl-:-KyPll8oHK4jHfKgT4pk:b-vnpav7) ([raw file](https://github.com/powershell/sharepointdsc/blob/81be4f8ab19f4a2885cf192e50c62051585c43e7/Modules/SharePointDsc/DSCResources/MSFT_SPManagedMetadataServiceApp/MSFT_SPManagedMetaDataServiceApp.psm1#L97)):*
<details><summary><i>Previously, ykuijs (Yorick Kuijs) wrote…</i></summary><blockquote>

Why are you generating the proxy name and not actually retrieving it, like we do it in SPSecureStoreServiceApp

            $serviceAppProxies = Get-SPServiceApplicationProxy -ErrorAction SilentlyContinue
            if ($null -ne $serviceAppProxies)
            {
                $serviceAppProxy = $serviceAppProxies | Where-Object -FilterScript { 
                    $serviceApp.IsConnected($_)
                }
                if ($null -ne $serviceAppProxy) 
                { 
                    $proxyName = $serviceAppProxy.Name
                }
            }
</blockquote></details>

I'm already using this snipped in Lines 82 to 95. I think the part from line 97 to 100 is obsolete. This would only be used if there is no proxy available at all. This would still make no sense :) I will delete these lines of code.

---


*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/666)*

andikrueger:
Removing the three lines was an horrible idea. I'm currently reworking the mock objects, as they are very different in the various cases... The Function IsConnected of an SPServiceApplication is now mandatory to the retun object of Get-SPServiceApplication.

ykuijs:
Reviewed 1 of 2 files at r7, 2 of 2 files at r8.

---



*Comments from [Reviewable](https://reviewable.io:443/reviews/powershell/sharepointdsc/666)*

Issue #804, opened 2018-05-23
SPUserProfileSyncConnection: Connection get's recreated every time if the name contains a dot


cvollmann:
**Details of the scenario you try and problem that is occurring:**

We're adding multiple sync connections to the UPS.
If the name for the connection contains a ".", this connections is recreated every time you run the script.
We were able to "fix" it, by replacing the dots with hyphens (see code below).

The connection was already beeing named with the hyphen in SP, even without replacing the dot. So if the forest is "ads.foobar.de" it shows up as: Name "ads-foobar-de" and forest "ads.foobar.de".

    $SharePointConfigData.Services.UserProfiles.SyncConnections | ForEach-Object -Process {
        SPUserProfileSyncConnection $_.Forest
        {
            Name                  = $_.Forest.Replace(".", "-")
            Forest                = $_.Forest
            UserProfileService    = $SharePointConfigData.Services.UserProfiles.Name
            IncludedOUs           = $_.IncludedOUs
            ConnectionType        = $connectionType
            ConnectionCredentials = $UpsSyncAccount
            UseSSL                = $false

            DependsOn             = "[SPUserProfileServiceApp]UserProfileServiceApp"
            PsDscRunAsCredential  = $SpSetupAccount
        }
    }

**Example from our configuration file:**

    SyncConnections  = @(
        @{
            Forest      = "adm.bla.blub.ads.foobar.de"
            IncludedOUs = @(
                "CN=MANAGED SERVICE ACCOUNTS,DC=ADM,DC=SWE,DC=baz,DC=ADS,DC=foobar,DC=DE",
                "CN=USERS,DC=ADM,DC=SWE,DC=baz,DC=ADS,DC=foobar,DC=DE",
                # <snip>
                "OU=TESTSERVERS,DC=ADM,DC=SWE,DC=baz,DC=ADS,DC=foobar,DC=DE"
            )
        },
        @{
            Forest      = "ADS.foobar.COM"
            IncludedOUs = @(
                "CN=USERS,DC=ADS,DC=foobar,DC=COM",
                # <snip>
                "OU=RO,DC=ADS,DC=foobar,DC=COM"
            )
        }
        # <snip>
    }

**Version of the Operating System and PowerShell the DSC Target Node is running:**
Windows Server 2012 R2, PowerShell 5.1

**Version of SharePoint that is used (e.g. SharePoint 2016):**
SharePoint 2016

**Version of the DSC module you're using:**
2.1.0.0

[...]

PR #814, opened 2018-06-07
Fixes for Connection Display Name

NikCharlebois:
Thanks for submitting a Pull Request to this project.
This template will help you create your pull request.

Please make sure you have read the [Contribution Guidelines](https://github.com/powershell/xSharePoint/wiki/Contributing%20to%20xSharePoint).

To aid community reviewers in reviewing and merging your PR, please take the time to run through this checklist:

**DELETE THIS LINE AND ABOVE**

[Replace this with a description of your pull request]

- [ ] Change details added to Unreleased section of changelog.md?
- [ ] New/changed code adheres to [Style Guidelines](https://github.com/PowerShell/DscResources/blob/master/StyleGuidelines.md)?

**DELETE THIS LINE AND BELOW**

Your contribution to this project is greatly appreciated.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/814)

### DIFF
--- a/Modules/SharePointDsc/DSCResources/MSFT_SPUserProfileSyncConnection/MSFT_SPUserProfileSyncConnection.psm1
+++ b/Modules/SharePointDsc/DSCResources/MSFT_SPUserProfileSyncConnection/MSFT_SPUserProfileSyncConnection.psm1
@@ -81,6 +81,7 @@ function Get-TargetResource
             UseDisabledFilter = $null
             ConnectionType = $null
             Force = $null
+            Ensure = "Absent"
         }
 
         if ($null -eq $ups)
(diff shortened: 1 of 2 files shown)


ykuijs:
You PR basically reverts the changes in my previous PR, which was a fix for issue #804 . There the resource tried to create the connection over and over again. If we would like to enable to use the dot in the name, we should also counter for the scenario where new connections will be created by DSC.

NikCharlebois:
@Yorick yes, I gave some thoughts around that at first, but really, how can we tell if the connection was created by hand or by DSC? Even if I create it by hand, I can decide to call it "contoso.com" or "contoso-com". Should the Set-TargetResource create the connection, then performs a check against the created name and the name passed as a parameter and rename it if they don't match?

So for example, if the connection needs to be "contoso.com", DSC will create it as "contoso-com", and then Set-TargetResource would rename it back to "contoso.com"? This would handle the Set, and the Get would simply check for an exact match based on the name received.

ykuijs:
The issue is that the "Add-SPProfileSyncConnection" cmdlet does not have an Name parameter. Internally it uses the Forest parameter, replacing the dot with a dash (I verified by looking at the code). So OOTB SharePoint acts differently when you create a connection from the GUI (you can specify a name) and by PowerShell (you cannot specify a name).

I think we should cater for both situations. Maybe by first checking if the specified name exist and if it does not, check for the Forest name where the dot is replaced by dash. But this does mean there could be a situation where you have two connections: One with dots and one with dashes. This resource is then unable to handle both situations......

ykuijs:
:lgtm:

---

Reviewed 2 of 2 files at r3.
Review status: :shipit: complete! all files reviewed, all discussions resolved

---



*Comments from [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/814#-:-LErtq2PCwCjYlMh0FRF:bnfp4nl)*

PR #836, opened 2018-07-11
Fix: SPInfoPathFormsServiceConfig - MaxSizeOfUserFormState parameter could never be set

cvollmann:
This PR fixes a bug for setting the "MaxSizeOfUserFormState" for InfoPath Forms Services. If you've tried to set this parameter it was always throwing an error and the default value was (still) used.

The assignment was basically "current value = (current value * 1024)" before.

Not sure if someone could change the test too (not familiar with testing these things, sorry).
The test uses the default value of this service application. So even if the value wasn't changed with the Set method - the tests could never reveal that because of the default value that's tested against.

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/836)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,6 +19,8 @@
   * Fixed code styling issues
 * SPUserProfileServiceApp
   * Fixed code styling issues
+* SPInfoPathFormsServiceConfig
+  * Fixed issue with trying to set the MaxSizeOfUserFormState parameter
 
 ## 2.3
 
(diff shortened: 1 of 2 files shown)


NikCharlebois:
@cvollmann Thanks for catching that. The test does not require any modification. I'll review this PR once AppVeyor finishes. Cheers

ykuijs:
Merging this PR since the requested change has been applied.

PR #892, opened 2018-09-23
SPServiceAppSecurity: Added support for local farm token and fixed multiple issues

### LINKED ISSUE #881: MSFT_SPServiceAppSecurity: Not possible to specify Local Farm
There are no way to specify Local Farm when specifying permissions for service applications. This is specified as the ID (Guid) of the local farm.

N/A
Add token to specify Local Farm as username in SPServiceAppSecurityEntry and convert it to $(Get-SPFarm).Id when executing the resource.

        $members = @()
        $members += MSFT_SPServiceAppSecurityEntry {
                                Username    = "Local Farm"
                                AccessLevel = "Full Control"
                            }
        SPServiceAppSecurity UserProfileServiceSecurity
        {
            ServiceAppName       = "Managed Metadata Service"
            SecurityType         = "SharingPermissions"
            Members     = $members
            PsDscRunAsCredential = $SPSetupCredential 
        }

Windows Server 2012 R2

#### Version of SharePoint that is used (e.g. SharePoint 2016)
SharePoint 2016

5.1
2.5.0.0

jensotto:
Adds support for local farm token.
Fixed many issues in the resource that prevented it from working as expected.
Some of them are:
Empty array specified for Members never worked.
Test-TargetResource never worked for MembersToExclude .
Set-TargetResource never worked for MembersToExclude.
Set-TargetResource did not work as expected when Members was specified. Members not specified was either not removed or an unexpected member was removed.
Many unit test contexts tested multiple situations that leaked test data between assertions that resulted in false assertions and undiscovered bugs.

A general observation is that I'm not sure if I can trust the existing unit tests in general as I struggled big time with issues that were not discovered by the unit tests, even though the unit tests stated that the specific use case was tested. This may also be true for other resources. The issues I had was only discovered by executing the resource on a live system. It might be that the quality of the unit tests are not good enough or that the unit tests tries to test too many things in a single context.

Regardless, I've tested all possible combinations (hopefully) for this resource on a live SharePoint 2013 installation.

- Fixes #881 

#### Task list
- [x] Resource documentation added/updated in README.md in the resource folder.
- [ ] Resource parameter descriptions added/updated in schema.mof
      and comment-based help.
- [x] Comment-based help added/updated.
- [x] Examples appropriately added/updated.
- [ ] Integration tests added/updated (where possible). See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/892)


jensotto:
I just realised that the farm id claim has to be checked for the local farm id and not only if it is a farm id claim type. The service application could be shared with other farms.
This also leads to a new bug. Currently the resource does not support anything other than a user account or a security group except for the local farm token that I just added. I have no farm set up to test sharing of service applications, but I think the claim should be similar to the local farm claim except for the farm id part.
This is however a defect that the resource already has, and this PR does nothing to change that.
Maybe we need to introduce another token to handle remote farms. Any ideas?

jensotto:
@ykuijs I don't understand why the code coverage shows only 83%. Locally it reports about 92% coverage.

johlju:
@jensotto I think it's the relative scope, see [Codecov Delta](https://docs.codecov.io/docs/codecov-delta), if you click on the pattern (green square), you see 83% in the relative scope, so think that number is representing that. The important row that I look at is `+ Coverage     92%     92%   +<1%`, and the hits an misses, it shows that all code is covered (as you also see locally).

jensotto:
PR is ready to be processed now

jensotto:
@ykuijs There seems to be problems with Reviewable. It does not allow me to log in anymore. Regarding your comments the $null checks could technically be added for MembersToInclude and MembersToExclude, but it makes no sense to allow those two properties to specify an empty array. Only for Members it makes sense so that all existing members can be removed.
What problem are you refering to? The unit tests?

ykuijs:
I am referring to the issue you mentioned in your first reply to this PR, about adding permissions to a remote farm

jensotto:
Yes, of course. I'll do that.

ykuijs:
You are right that an empty array only makes sense for the Members parameter, however the check you updated is checking that at least one of the three members parameters is specified. And to make sure the code is consistent, I prefer all three checks to use "$null -eq [parameter_name]"

jensotto:
@ykuijs Done.

ykuijs:
@jensotto You are on a roll!! Thanks mate!

PR #1018, opened 2019-01-30
SPWebAppPeoplePickerSettings: Converted access account password to secure string

### LINKED ISSUE #1019: SPWebAppPeoplePickerSettings: Unable to set AccessAccount
When specifying AccessAccount the resource fails to set the password.

PowerShell DSC resource MSFT_SPWebAppPeoplePickerSettings  failed to execute Set-TargetResource functionality with erro
r message: Cannot convert argument "password", with value: "xxxx", for "SetPassword" to type "System.Security.Sec
ureString": "Cannot convert the "xxxx" value of type "System.String" to type "System.Security.SecureString"."
    + CategoryInfo          : InvalidOperation: (:) [], CimException
    + FullyQualifiedErrorId : ProviderOperationExecutionFailure
    + PSComputerName        : xxxx
The password needs to be converted to SecureString

```powershell
            SPWebAppPeoplePickerSettings PeoplePicker
            {
                WebAppUrl = $ConfigurationData.NonNodeData.Url
                ActiveDirectoryCustomFilter = ''
                ActiveDirectoryCustomQuery = $null
                ActiveDirectorySearchTimeout = 30
                OnlySearchWithinSiteCollection = $false
                SearchActiveDirectoryDomains = @(
                    MSFT_SPWebAppPPSearchDomain {
                        FQDN     = "domain1.com" 
                        IsForest = $true
                    }
                    MSFT_SPWebAppPPSearchDomain {
                        FQDN     = "domain2.com"
                        AccessAccount = $SPADSyncCredentials                        
                        IsForest = $true
                    }
                )
                PsDscRunAsCredential = $SPSetupCredential
            }
```

OsName               : Microsoft Windows Server 2016 Datacenter
OsOperatingSystemSKU : DatacenterServerEdition
OsArchitecture       : 64-bit
WindowsBuildLabEx    : 14393.2724.amd64fre.rs1_release.181231-1751
OsLanguage           : en-US
OsMuiLanguages       : {en-US}#### Version of SharePoint that is used (e.g. SharePoint 2016)

Name                           Value
----                           -----
PSVersion                      5.1.14393.2636
PSEdition                      Desktop
PSCompatibleVersions           {1.0, 2.0, 3.0, 4.0...}
BuildVersion                   10.0.14393.2636
CLRVersion                     4.0.30319.42000
WSManStackVersion              3.0
PSRemotingProtocolVersion      2.3
SerializationVersion           1.1.0.1

3.1.0.0

jensotto:
Converted password for access account to secure string. Previsouly the resource would fail setting the password and an exeption was thrown that printed the password in clear text.
I would expect the password to already be a secure string, but apparently that's not the case.

- Fixes #1019 

#### Task list
- [ ] Resource documentation added/updated in README.md in the resource folder.
- [ ] Resource parameter descriptions added/updated in schema.mof
      and comment-based help.
- [ ] Comment-based help added/updated.
- [ ] Examples appropriately added/updated.
- [ ] Integration tests added/updated (where possible). See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/1018)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -33,6 +33,10 @@
   * Fix user profile property mappings does not work
 * SPUserProfileSyncService
   * Fixed issue where failure to configure the sync service would not throw error
+* SPWebAppPeoplePickerSettings
+  * Converted password for access account to secure string. Previsouly
+    the resource would fail setting the password and an exeption was thrown that
+    printed the password in clear text.
 * SPWebAppPolicy
   * Fixed issue where parameter MembersToExclude did not work as expected
 * SPWorkflowService
--- a/Modules/SharePointDsc/DSCResources/MSFT_SPWebAppPeoplePickerSettings/MSFT_SPWebAppPeoplePickerSettings.psm1
+++ b/Modules/SharePointDsc/DSCResources/MSFT_SPWebAppPeoplePickerSettings/MSFT_SPWebAppPeoplePickerSettings.psm1
@@ -183,16 +183,20 @@ function Set-TargetResource
                     }
 
                     $adsearchobj.IsForest = $searchADDomain.IsForest
-                    $adsearchobj.LoginName = $searchADDomain.AccessAccount.UserName
 
-                    if([string]::IsNullOrEmpty($searchADDomain.AccessAccount.Password))
+                    if ($null -ne $searchADDomain.AccessAccount)
                     {
-                        $adsearchobj.SetPassword($null)
-                    }
-                    else
-                    {
-                        $accessAccountPassword = ConvertTo-SecureString $searchADDomain.AccessAccount.Password -AsPlainText -Force
-                        $adsearchobj.SetPassword($accessAccountPassword)
+                        $adsearchobj.LoginName = $searchADDomain.AccessAccount.UserName
+
+                        if([string]::IsNullOrEmpty($searchADDomain.AccessAccount.Password))
+                        {
+                            $adsearchobj.SetPassword($null)
+                        }
+                        else
+                        {
+                            $accessAccountPassword = ConvertTo-SecureString $searchADDomain.AccessAccount.Password -AsPlainText -Force
+                            $adsearchobj.SetPassword($accessAccountPassword)
+                        }
                     }
 
                     $wa.PeoplePickerSettings.SearchActiveDirectoryDomains.Add($adsearchobj)
--- a/Modules/SharePointDsc/DSCResources/MSFT_SPWebAppPeoplePickerSettings/MSFT_SPWebAppPeoplePickerSettings.psm1
+++ b/Modules/SharePointDsc/DSCResources/MSFT_SPWebAppPeoplePickerSettings/MSFT_SPWebAppPeoplePickerSettings.psm1
@@ -79,6 +79,7 @@ function Get-TargetResource
 
 function Set-TargetResource
 {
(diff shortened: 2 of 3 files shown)


jensotto:
@ykuijs I spoke too early regarding this one. I can't get the ConvertTo-SecureString cmdLet to work in a live system. Do you know why the original code does not work and any alternative approach to get it working? The error I get when using ConvertTo-SecureString is Cannot bind argument to parameter 'String' because it is null.

jensotto:
I figured it out. It's the if statement if($null -ne $prop) that always is false (AccessAccount is always present in the CimInstanceProperties even if not specified). I incorrectly assumed that this if statement would filter out missing search domains with access accounts not specified. I'll update the code soon.

jensotto:
@ykuijs I've updated the code now, and it works as expected. I implemented it so that the login and password still are set even if the AccessAccount is not specified. This way it's possible to clear out existing credentials. This is also how the code previously worked, except that setting credentials never actually worked.

ykuijs:
Will review once AppVeyor completes. In the meantime:
Can you also registered an issue for the issue that is fixed in this PR? That way we can track the issue properly.

ykuijs:
Investigated this issue and found the reason: The AccessAccount is handled as a CimInstance class instead of a PSCredential. This is causing the Password property to be plain text :-(

Since Pester is checking for the use of Plain text passwords, we have to disable that check in order to pass the tests:

Can you please add the following lines to the Set method, so the Plain Text check is not performed and the AppVeyor test completes successfully:
```
function Set-TargetResource
{
    # Ignoring this because the used AccessAccount stores the password in Plain Text
    [Diagnostics.CodeAnalysis.SuppressMessageAttribute("PSAvoidUsingConvertToSecureStringWithPlainText", "")]
```

ykuijs:
One last question: Since we will release v3.2 next Wednesday, would it be possible to complete this PR asap, so we can include it into that release?

jensotto:
I added the attribute with some minor adjustments to the justification message so that it does not sounds like that the password is stored in plain text in the MOF-file. Also used the Justification argument for the SuppressMessageAttribute instead of using a code comment.

Is there no way to fix this so that AccessAccount uses a SecurePassword?

jensotto:
___
*[Modules/SharePointDsc/DSCResources/MSFT_SPWebAppPeoplePickerSettings/MSFT_SPWebAppPeoplePickerSettings.psm1, line 185 at r2](https://reviewable.io/reviews/powershell/sharepointdsc/1018#-LXZahk821V46uSTs_l6:-LYQgQZO9OAiok2OQT9e:b-q4lrbx) ([raw file](https://github.com/powershell/sharepointdsc/blob/40d29f784d62178b5310acbbc99e9ba58ba0e460/Modules/SharePointDsc/DSCResources/MSFT_SPWebAppPeoplePickerSettings/MSFT_SPWebAppPeoplePickerSettings.psm1#L185)):*
<details><summary><i>Previously, ykuijs (Yorick Kuijs) wrote…</i></summary><blockquote>

Just tested this and in my case, $prop has a value when AccessAccount has a value.

Second: How can the issue you encountered be generated if $prop is always null? Then the specific code should never be called.
</blockquote></details>

$prop was never $null. It always had a value even if AccessAccount was not specified. That was why I got the issue. I assumed that the code would not execute if AccessAccount was not specified. This was not the case, and the code executed always leading to an exception when converting the password to SecureString.
So based on your testing you have found that $prop has a value which is the same as I saw, including when AccessAccount is not specified.

PR #1141, opened 2019-11-26
SPUserProfileServiceApp: Fix proxy group issue

### LINKED ISSUE #1140: SPUserProfileServiceApp: Add ability to correct issue with incorrectly linked proxy groups
A support case for a customer revealed that the UP service verifies an obsolete property which is not updated correctly. This can result in issues with the User Profile page:

"We’ve investigated it and found that when the User Profile service is looking up the proxy groups to find the correct MMS, it doesn’t follow what you configured in Central Administration. Instead it verifies an obsolete ServiceApplicationProxyGroup property of the UPA, which is always pointing to the default proxy group.

It seems that property doesn’t get updated for any Service Application when you modify the associations in Central Administration, and it appears it can store only one association. (while in central administration a Service Application can be added to multiple proxy groups).

The good news is, that property be configured using PowerShell. The commands would look similar too:"

```PowerShell
Get-SPServiceApplication | select id,displayname
$upa = Get-SPServiceApplication -Identity <guid1>
$upa.ServiceApplicationProxyGroup
$g = Get-SPServiceApplicationProxyGroup 

$upa.ServiceApplicationProxyGroup = $g[z]
$upa.Update()
```
Where z is the id of the correct proxy group

N/A

Implement a new parameter called UpdateProxyGroup. The Get method would check if the configured proxy group matches the ServiceApplicationProxyGroup property and return False if this is not the case.

If the property is set to True, the Set method would update the ServiceApplicationProxyGroup property to the proxy group to which the service app is linked.

N/A

Any

#### Version of SharePoint that is used (e.g. SharePoint 2016)
Any

Any

dev

ykuijs:
A support case for a customer revealed that the UP service verifies an obsolete property which is not updated correctly. This can result in issues with the User Profile page. This PR fixes this issue.

- Fixes #1140 

#### Task list
- [x] Resource documentation added/updated in README.md in the resource folder.
- [x] Resource parameter descriptions added/updated in schema.mof
      and comment-based help.
- [x] Comment-based help added/updated.
- [x] Localization strings added/updated in all localization files as appropriate.
- [x] Examples appropriately added/updated.
- [x] Integration tests added/updated (where possible). See [DSC Resource Testing Guidelines](https://github.com/PowerShell/DscResources/blob/master/TestsGuidelines.md).

---
This change is [Reviewable](https://reviewable.io/reviews/powershell/sharepointdsc/1141)

### DIFF
--- a/.github/CONTRIBUTING.md
+++ b/.github/CONTRIBUTING.md
@@ -1,1 +1,1 @@
-If you'd like to contribute to this project, please review the [Contribution Guidelines](https://github.com/powershell/SharePointDsc/wiki/Contributing%20to%20SharePointDsc).
+If you'd like to contribute to this project, please review the [Contribution Guidelines](https://github.com/dsccommunity/SharePointDsc/wiki/Contributing%20to%20SharePointDsc).
--- a/.github/PULL_REQUEST_TEMPLATE.md
+++ b/.github/PULL_REQUEST_TEMPLATE.md
@@ -2,7 +2,7 @@
     Thanks for submitting a Pull Request (PR) to this project.
     Your contribution to this project is greatly appreciated!
 
-    Please make sure you have read the [Contribution Guidelines](https://github.com/PowerShell/SharePointDsc/wiki/Contributing%20to%20SharePointDsc).
+    Please make sure you have read the [Contribution Guidelines](https://github.com/dsccommunity/SharePointDsc/wiki/Contributing%20to%20SharePointDsc).
 
     Please prefix the PR title with the resource name,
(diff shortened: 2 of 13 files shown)


ykuijs:
@NikCharlebois If you have a moment.....

ykuijs:
With the move to the Dsc Community org, we are no longer allowed to run AppVeyor builds longer than one hour. Since our tests run longer, AppVeyor fails.

I am now checking with Gael to resolve this issue by moving to Azure DevOps. Unfortunately this means that tests are failing in the meantime.

PR #1473, opened 2026-05-05
Avoid random error with Write-EventLog in Windows Server 2025

Yvand:

Randomly (about 15-20%), deployments fail in Windows Server 2025 because of error "The registry key for the log "SPDsc" for source "MSFT_SPFarm" could not be opened", when running cmdlet `Write-EventLog` in function `Add-SPDscEvent`.

Please note:
- This error is not easy to repro, it works fine most of the time, but somehow it happens from time to time with Windows Server 2025
- This change is totally safe, since it only complements the existing try/catch (which does not trap cmdlet errors). Beside, if the logging fails, we really do not want this to interrupt the deployment.

#### Task list

- [x] Added an entry to the change log under the Unreleased section of the file CHANGELOG.md.
      Entry should say what was changed and how that affects users (if applicable), and
      reference the issue being resolved (if applicable).
- [ ] Resource documentation added/updated in README.md.
- [ ] Resource parameter descriptions added/updated in README.md, schema.mof and comment-based
      help.
- [ ] Comment-based help added/updated.
- [ ] Examples appropriately added/updated.
- [ ] Unit tests added/updated. See [DSC Community Testing Guidelines](https://dsccommunity.org/guidelines/testing-guidelines).
- [ ] Integration tests added/updated (where possible). See [DSC Community Testing Guidelines](https://dsccommunity.org/guidelines/testing-guidelines).
- [x] New/changed code adheres to [DSC Community Style Guidelines](https://dsccommunity.org/styleguidelines).

- - -
This change is [Reviewable](https://reviewable.io/reviews/dsccommunity/SharePointDsc/1473)

### DIFF
--- a/SharePointDsc/Modules/SharePointDsc.Util/SharePointDsc.Util.psm1
+++ b/SharePointDsc/Modules/SharePointDsc.Util/SharePointDsc.Util.psm1
@@ -48,7 +48,7 @@ function Add-SPDscEvent
     try
     {
         Write-EventLog -LogName $LogName -Source $Source `
-            -EventId $EventID -Message $Message -EntryType $EntryType -ErrorAction SilentlyContinue
+            -EventId $EventID -Message $Message -EntryType $EntryType -ErrorAction Stop
     }
     catch
     {
--- a/SharePointDsc/Modules/SharePointDsc.Util/SharePointDsc.Util.psm1
+++ b/SharePointDsc/Modules/SharePointDsc.Util/SharePointDsc.Util.psm1
@@ -48,7 +48,7 @@ function Add-SPDscEvent
     try
     {
         Write-EventLog -LogName $LogName -Source $Source `
-            -EventId $EventID -Message $Message -EntryType $EntryType
+            -EventId $EventID -Message $Message -EntryType $EntryType -ErrorAction SilentlyContinue
     }
     catch
     {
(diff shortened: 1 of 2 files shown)


SharePointDsc/Modules/SharePointDsc.Util/SharePointDsc.Util.psm1
```diff
@@ -48,7 +48,7 @@ function Add-SPDscEvent
     try
     {
         Write-EventLog -LogName $LogName -Source $Source `
-            -EventId $EventID -Message $Message -EntryType $EntryType
+            -EventId $EventID -Message $Message -EntryType $EntryType -ErrorAction SilentlyContinue
```
ykuijs: By adding -ErrorAction SilentlyContinue, the try/catch basically becomes useless. The Write-EventLog never throws an exception because of that.

The strange thing is that the catch block shows a verbose message, so the code should never cause a blocking issue. The only thing I can think of is that it shows an error instead of throwing an exception. That would mean the cmdlet does throw the error, but since catch blocks only catch exceptions the catch block would never be hit.

In order to test this, could you change the above change to `-ErrorAction Stop`? That forces the cmdlet to throw an exception when an error occurs and the catch block would be triggered.

ykuijs (changes requested):
Do not think this change is the solution to this issue

SharePointDsc/Modules/SharePointDsc.Util/SharePointDsc.Util.psm1
Yvand: This is a great suggestion, I tested it and it is the best one indeed, thank you!

ykuijs (approved):
LGTM

